PR #39, opened 2023-12-11
fix: minor syntax error in example code

bstncartwright:
Fix of a small syntax error in the example code 😄


maaslalani:
Thank you so much! ❤️

PR #40, opened 2023-12-11
fix max width logic

ardnew:
To ensure the form is no greater than maxWidth, we need to take the minimum among the desired width and maxWidth.

Taking the maximum among these values effectively changes the logic to ensure the form is no less than maxWidth (in which case, it should probably be called minWidth).


maaslalani:
Good catch @ardnew! Thank you for the PR!

PR #43, opened 2023-12-12
feat(theme): add `Catppuccin`

sgoudham:
Hey :wave:
First of all, this is a lovely project, thanks for creating it!

I'm just following up on @maaslalani's discord message, and adding Catppuccin alongside the default themes.

**P.S** I'm not a go programmer™ so please let me know if I've changed something that I shouldn't have!

### DIFF
--- a/theme.go
+++ b/theme.go
@@ -1,6 +1,7 @@
 package huh
 
 import (
+	catppuccin "github.com/catppuccin/go"
 	"github.com/charmbracelet/bubbles/help"
 	"github.com/charmbracelet/lipgloss"
 )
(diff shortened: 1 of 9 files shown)


sgoudham:
One thing I wanted to clarify, can the colours for the actions (see screenshot below) be modified?

image

I wasn't able to find obvious keys for the `next` and `back` text so it's quite harsh on Catppuccin's darker flavours.

theme.go
```diff
@@ -270,3 +271,66 @@ func ThemeBase16() *Theme {
...
+	return themeCatppuccin(catppuccin.Frappe)
+}
+
+// ThemeCatppuccinMacchiato returns a new theme based on the Macchiato flavor of the Catppuccin color scheme.
+func ThemeCatppuccinMacchiato() *Theme {
+	return themeCatppuccin(catppuccin.Macchiato)
+}
+
+// ThemeCatppuccinMocha returns a new theme based on the Mocha flavor of the Catppuccin color scheme.
+func ThemeCatppuccinMocha() *Theme {
```
sgoudham, line 334: I've added the 4 flavours of Catppuccin here, without the use of `lipgloss.AdaptiveColor` as we have a light and dark theme. 

However, there's a case to be made where instead of doing all 4, we can have one single `Catppuccin` theme that uses `lipgloss.AdaptiveColor` to switch between our light flavour (Latte) and our darkest flavour (Mocha) accordingly. 

Now that I'm typing this out, I think that might actually be a nicer approach

meowgorithm:
> One thing I wanted to clarify, can the colours for the actions (see screenshot below) be modified?
> 
> I wasn't able to find obvious keys for the `next` and `back` text so it's quite harsh on Catppuccin's darker flavours.

Hey, @sgoudham! You sure can . It's `Help` here on `FieldStyles`:

https://pkg.go.dev/github.com/charmbracelet/huh#FieldStyles

And then the help definition is:

https://pkg.go.dev/github.com/charmbracelet/bubbles@v0.16.1/help#Styles

Let us know if you have more questions. We’re really excited about this Catpuccin theme!

theme.go
maaslalani: Thanks so much for going with this approach! I really like it!

maaslalani:
Thanks so much! This looks perfect!

meowgorithm:
Holy moly that looks good, @sgoudham. Nice work!

sgoudham:
Thank you both for making my first contribution into Charm so easy and fun! :heart:

maaslalani:
Thank _you_ for making such an awesome contribution!

We just released `v0.2.0` to include the new theme: https://github.com/charmbracelet/huh/releases/tag/v0.2.0

PR #44, opened 2023-12-12
fix: set help style

jolheiser:
This should set the help style to whatever is set as the `Focused.Help` style. 
It may make more sense to have a dedicated help style, but the other themes seem to follow a pattern of `Focused` being primary and `Blurred` copying and modifying.


meowgorithm:
@jolheiser Good catch, and you’re totally right about the dedicated help style. We’ll make that change accordingly.

maaslalani:
Hey @jolheiser, thank you for the PR!

For a more correct solution we will put the `Help` styles on the theme itself instead of embedding within the Focused/Blurred of a field. Since the theme help applies to the group/form not a field.


```go
// Currently
theme.Focused.Help.ShortKey = lipgloss.NewStyle() //...

// Suggested
theme.Help.ShortKey = lipgloss.NewStyle() //...
```

jolheiser:
I think that sounds great and agree it would be much cleaner. I'll close this one and let you hammer it out, unless you would like me to amend this PR. :slightly_smiling_face:

maaslalani (approved):
Nice @jolheiser, this is absolutely perfect! Thank you so much!

PR #48, opened 2023-12-13
feat: dependabot, bubbletea update, go versions

caarlos0:
this will allow to run huh on go 1.19, latest bubbletea version, and adds the dependabot configuration as well.

also updates the build workflow to build and test all modules with proper go version (1.19 instead of 1.18)


maaslalani:
Thanks so much @caarlos0!

PR #57, opened 2023-12-14
docs: fix repeated word "mode"

vitor-mariano:
The word "mode" is repeated twice in readme.

CleanShot 2023-12-14 at 09 30 32@2x


maaslalani (approved):
Thanks so much for the PR @vitor-mariano ❤️

PR #58, opened 2023-12-14
feat(options): simple options (key = value)

phm07:
This PR adds the method `NewSimpleOption(key string)`, which is a more idiomatic way to create options that have the same key and value.

The main advantages of such a method are:
- When passing an expression, there is no need to introduce a temporary variable or to write it twice
- When passing a string literal, there is no risk of there being a typo so that key != value
- Serves the DRY principle

IMO this also makes the code easier to read, as it is not instantly obvious why there are 2 parameters with the same value when using `NewOption`. I also think this use case is common enough to warrant a dedicated method for it. Let me know what you think of this suggestion 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,13 +205,13 @@ Prompt the user to select multiple (zero or more) options from a list.
 ```go
 huh.NewMultiSelect[string]().
     Options(
-        huh.NewOption("Lettuce", "Lettuce").Selected(true),
-        huh.NewOption("Tomatoes", "Tomatoes").Selected(true),
-        huh.NewOption("Charm Sauce", "Charm Sauce"),
-        huh.NewOption("Jalapeños", "Jalapeños"),
-        huh.NewOption("Cheese", "Cheese"),
-        huh.NewOption("Vegan Cheese", "Vegan Cheese"),
-        huh.NewOption("Nutella", "Nutella"),
+        huh.NewSimpleOption("Lettuce").Selected(true),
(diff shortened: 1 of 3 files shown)


maaslalani:
Hey @pauhull, thank you for the PR. Great observation on the more simple use cases.

We have introduced a different method called `NewOptions` for this exact purpose (when `key` = `value`) which would be more concise than using `NewSimpleOption`:

```go
huh.NewOptions("Lettuce", "Tomatoes", "Jalepeños", "Cheese", "Vegan Cheese", "Nutella")
```

Does this solve your use case?

pauhull:
In order to use that method you have to have a string slice first. If you have a slice of some other type that you want to map to options you would need to create a string slice first. Additionally, the `NewSimpleOption` would allow to change further option properties. For example, this is my use case:
```go
var opts []huh.Option[string]
for _, ctx := range cli.Config.Contexts {
	opts = append(opts, huh.NewSimpleOption(ctx.Name).Selected(ctx == cli.Config.ActiveContext))
}
```

maaslalani:
Hey @phm07, thanks again for your contribution!

I like the idea and I think it does solve a valid use case. However, I would like to keep the API surface area as small as possible so that there is less to learn. It's also easy enough for the user to define their own `newSimpleOption` function to accomplish this.

I will close this PR for now since I'm doing some clean up but please feel free to re-open or create a new issue if this is still something you need!

Issue #54, opened 2023-12-14
No wrapping on answer exceeding the terminal length


Iilun:
**Describe the bug**
When using a single line input, if the answer length exceeds the terminal length, the cursor and the rest of the answer disappear.

**To Reproduce**
. Create an input 
. Input a very long answer 

**Expected behavior**
The exceeding characters are wrapped back to the next line.

**Screenshots**

https://github.com/charmbracelet/huh/assets/63926901/63b22dc0-77c0-438c-9bc1-cc6164ce4629

In this example the terminal stops at the right of the recording. I input only a's then delete some to come back in the frame.

**Desktop (please complete the following information):**
 - OS: macOS iTerm2 and Windows Powershell

**Additional context**
This may be an intended behavior but it seems not really user friendly

meowgorithm:
Hi again. Just a note that this was fixed in #60 and is now available in v0.2.2.

Iilun:
I tested it, its perfect, thanks !

PR #60, opened 2023-12-14
Set Width Correctly

maaslalani:
This PR sets the width of the `textinput` and `textarea` bubbles.

It calculates the widths from the form and horizontal frames. It also changes
the default width from 80 to the default size of the terminal with auto-resizing
if the user does not specify a Width.


maaslalani:
Fixes #54

Issue #55, opened 2023-12-14
Add the possibility to prefill the answer


Iilun:
**Is your feature request related to a problem? Please describe.**
When creating prompts for editing a file for example, some values may already have been filled previously. Instead of only showing it as placeholder and needing the user to retype it, it would be great if it was there and editable.

**Describe the solution you'd like**
An option on input to add a string as a default value that will be prefilled in the input

maaslalani:
Hey @Iilun, this is possible if you set the variable to the default value you want:

```go
value := "default value"

huh.NewInput().Value(&value).Run()
```

The field will be pre-filled with the value ("default value")

vitor-mariano:
@maaslalani All the fields can be prefilled with `Value(&value)`, except `Select` and `MultiSelect`. I know I can achieve the same behavior with `huh.NewOption(...).Selected(true)`, but not with `huh.NewOptions`. Is this on purpose? Using `Value(&value)` would be interesting in scenarios when editing existing data and keeping consistency with other field types.

If this behavior is not intended, I'd love to open a PR fixing it!

vitor-mariano:
@maaslalani before moving forward with the fix, what should happen if I have both value and `Selected(true)`? Eg

``` go
huh.NewMultiSelect().
    Options(
        huh.NewOption("lettuce", "lettuce"),
        huh.NewOption("tomatoes", "tomatoes").Selected(true),
        huh.NewOption("cheese", "cheese").Selected(true),
    ).
    Value(&toppings).
    Run()
```

maaslalani:
I think the correct solution here would be to take the union of the two defaults. In this case all three would be selected as lettuce is in the toppings and the other two are selected.

PR #62, opened 2023-12-14
fix: prefill select and multiselect with value

vitor-mariano:
Fixes https://github.com/charmbracelet/huh/issues/55

Dev notes: I had to change the value generic types to `comparable` to make the comparison between `option.Value` and the received values possible. Also, it seems more correct.

### DIFF
--- a/field_multiselect.go
+++ b/field_multiselect.go
@@ -11,7 +11,7 @@ import (
 )
 
 // MultiSelect is a form multi-select field.
-type MultiSelect[T any] struct {
+type MultiSelect[T comparable] struct {
 	value *[]T
 	key   string
 
@@ -38,7 +38,7 @@ type MultiSelect[T any] struct {
 }
 
 // NewMultiSelect returns a new multi-select field.
-func NewMultiSelect[T any]() *MultiSelect[T] {
+func NewMultiSelect[T comparable]() *MultiSelect[T] {
 	return &MultiSelect[T]{
 		options:  []Option[T]{},
 		value:    new([]T),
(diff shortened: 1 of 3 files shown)


field_multiselect.go
```diff
@@ -49,6 +49,14 @@ func NewMultiSelect[T any]() *MultiSelect[T] {
 // Value sets the value of the multi-select field.
 func (m *MultiSelect[T]) Value(value *[]T) *MultiSelect[T] {
 	m.value = value
```
maaslalani, line 51: Can `nil` be passed to `Value`? I think we should have a `nil` check here.

maaslalani:
I think we'll also have to perform this loop when passing options. The reason being we should avoid ordering of the values / options being set:

That is, the following two should be equivalent:

```
huh.NewMultiSelect().Options(options).Value(&value)
// should be equivalent to...
huh.NewMultiSelect().Value(&value).Options(options)
```

maaslalani (commented):
Looks great, thanks so much @vitor-mariano. Just a few comments.

vitor-mariano:
> I think we'll also have to perform this loop when passing options. The reason being we should avoid ordering of the values / options being set:
> 
> That is, the following two should be equivalent:
> 
> ```go
> huh.NewMultiSelect().Options(options).Value(&value)
> // should be equivalent to...
> huh.NewMultiSelect().Value(&value).Options(options)
> ```

Good catch! Just fixed it.

maaslalani:
Thanks so much @vitor-mariano! Really appreciate this improvement! ❤️

Issue #63, opened 2023-12-16
GetValue does not work for Text inputs


farant:
**Describe the bug**
I'm integrating Huh with a Bubbletea app and it seems like m.form.GetValue("field") does not work for Text fields

**To Reproduce**
```go
type Model struct {
    form *huh.Form // huh.Form is just a tea.Model
}

func NewModel() Model {
    return Model{
        form: huh.NewForm(
            huh.NewGroup(
	        huh.NewText().
		    Key("question").
		    Title("What's your question?"),
            ),
        )
    }
}

func (m Model) Init() tea.Cmd {
    return m.form.Init()
}

func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
    // ...

    form, cmd := m.form.Update(msg)
    if f, ok := form.(*huh.Form); ok {
        m.form = f
    }

    return m, cmd
}

func (m Model) View() string {
    if m.form.State == huh.StateCompleted {
        question := m.form.GetString("question")
        return fmt.Sprintf("You selected: %s", question)
    }
    return m.form.View()
}
```

**Expected behavior**
question should contain the value of the Text input once it's completed. It is empty. If I change the code to be `huh.NewInput(...)` getting the value works.

Thanks!

PR #64, opened 2023-12-16
fix: text value

vitor-mariano:
Fixes https://github.com/charmbracelet/huh/issues/63

`text` value property wasn't being updated from `textarea` value.

**before**
CleanShot 2023-12-16 at 16 58 23

**after**
CleanShot 2023-12-16 at 16 58 57


maaslalani (approved):

maaslalani:
Thanks so much for the fix @vitor-mariano! Really appreciated ❤️

PR #71, opened 2023-12-18
feat: scrollable multi-select

### LINKED ISSUE #67: Long multi-select cuts off items
**Describe the bug**
If you create e.g. a NewMultiSelect with a lot of items than it happens that some items are not visible (depending on the terminal hright

**To Reproduce**

```
package main

import (
	"strconv"

)

func main() {
	result := []string{}
	list := []huh.Option[string]{}
	for i := 0; i < 30; i++ {
		item := "item-" + strconv.Itoa(i)
		list = append(list, huh.NewOption(item, item))
	}
	sel := huh.NewMultiSelect[string]().Options(list...).Value(&result)
	form := huh.NewForm(
		huh.NewGroup(sel).Description("Long List"),
	)
	err := form.Run()
	if err != nil {
		panic(err)
	}
}
```

Run in a terminal which can show less then 30 lines.

**Expected behavior**
List should be scrollable 

**Screenshots**

Screenshot 2023-12-18 at 09 41 59

 - OSX, verson: v0.2.1

meowgorithm:
This adds a `Height(int)` setter to `MultiSelect` prompts.

```go
multiSelect := huh.NewMultiSelect[int]().
	Options(ops...).
	Height(10)
```

* When `Height <= 0` the multi-select auto-sizes itself to the number of fields (as it does currently)
* When `Height > 0` the multi-select will behave like a viewport and the visible options will scroll with the cursor

Note that the height here refers to the total height of the multi-select itself including the height of title and the description (if set) however the number of options visible will never be less than 1.

Here’s an example of it in use:

https://github.com/charmbracelet/huh/assets/25087/5c5a9cbe-1eb1-4066-bb4c-d6c69068bb77

And here's what that code looks like:

```go
package main

import (
	"fmt"
	"os"

)

func main() {
	const height = 10

	opts := make([]huh.Option[int], 50)
	for i := 0; i < len(opts); i++ {
		opts[i] = huh.Option[int]{
			Key:   fmt.Sprintf("Option %d", i+1),
			Value: i,
		}
	}

	prompt := huh.NewMultiSelect[int]().
		Title("Choose something, will ya?").
		Description(fmt.Sprintf("This prompt is %d cells high and has %d choices", height, len(opts))).
		Options(opts...).
		Height(height)

	if err := prompt.Run(); err != nil {
		fmt.Println("Uh oh: ", err)
		os.Exit(1)
	}
}
```

Closes #67.

### DIFF
--- a/field_multiselect.go
+++ b/field_multiselect.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/charmbracelet/bubbles/key"
+	"github.com/charmbracelet/bubbles/viewport"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/huh/accessibility"
 	"github.com/charmbracelet/lipgloss"
@@ -21,14 +22,16 @@ type MultiSelect[T comparable] struct {
 	options     []Option[T]
 	filterable  bool
 	limit       int
+	height      int
 
 	// error handling
 	validate func([]T) error
 	err      error
 
 	// state
-	cursor  int
-	focused bool
+	cursor   int
+	focused  bool
+	viewport viewport.Model
 
 	// options
 	width      int
@@ -94,6 +97,7 @@ func (m *MultiSelect[T]) Options(options ...Option[T]) *MultiSelect[T] {
 		}
 	}
 	m.options = options
+	m.updateViewportHeight()
 	return m
 }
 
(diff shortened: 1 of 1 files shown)


tetienne:
Thx for implementing this. Do you think it will take a lot of effort to port to Select too?

chilicat:
Looks great! The list would need some kind of indicator that more items are available. Would it make sense to show this "paging dots"?

meowgorithm:
> @tetienne: Thx for implementing this. Do you think it will take a lot of effort to port to Select too?

Easy enough, see #76.

> @chilicat: Looks great! The list would need some kind of indicator that more items are available. Would it make sense to show this "paging dots"?

Agreed, though that will require some updates to themes as well, so we should examine that one in a separate pass.

PR #75, opened 2023-12-19
docs(example): simple example showing conditional options

meowgorithm:
How we might go about officially something like this is a larger topic, but for now here's an example on how one can get started with forms that change based on user choices.

### DIFF
--- a/examples/conditional/main.go
+++ b/examples/conditional/main.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/charmbracelet/huh"
+)
+
+type consumable int
+
+const (
+	fruits consumable = iota
+	vegetables
+	drinks
+)
+
+func (c consumable) String() string {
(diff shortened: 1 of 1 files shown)


safaci2000:
Thank you for this, it's almost exactly what I was looking for.  Would also love a way to integrate multiple forms with bubble tea.  I tried to have an array of forms to have dynamic form behavior and was not able to get it working properly.

meowgorithm:
> Thank you for this, it's almost exactly what I was looking for. Would also love a way to integrate multiple forms with bubble tea. I tried to have an array of forms to have dynamic form behavior and was not able to get it working properly.

Totally possible. If you want to open a discussion in the meantime with some (simple) source code would be happy to offer some help.

meowgorithm:
Note: this PR should be refactored to use @caarlos0’s excellent `HideFunc`.

maaslalani:
@meowgorithm, refactored this PR to use `WithHideFunc` makes it so the user can go back and make a different selection which is great!

PR #80, opened 2023-12-21
fix: Select and MultiSelect height when unfocused

vitor-mariano:
When `Select` or `MultiSelect` had a defined height for scrolling (kudos @meowgorithm!) and wasn't the first field in the group, the height was ignored and re-rendered only when focused.

bug

Since the height is evaluated after the theming step, `updateViewportHeight` needs to be called in `WithTheme` and `Height` functions.

fix


maaslalani (approved):
Thanks so much @vitor-mariano, really appreciate this fix! ❤️

PR #84, opened 2023-12-22
fix: Select viewport on filtering

vitor-mariano:
When you move the `Select` cursor to the bottom and filter for something, the whole viewport moves beyond the top.

bug

With this fix, I clamped the offset to the view on both top and bottom.

fix


maaslalani (approved):

maaslalani:
Thank you so much for all these wonderful PRs, @vitor-mariano!

PR #94, opened 2024-01-04
Reduce binary size

maaslalani:
Reduce binary size by removing some dependencies.


maaslalani:
Before:

```
33M     ./burger
```

After:

```
4.5M    ./burger
```

PR #96, opened 2024-01-10
feat: Titlef and Descriptionf

caarlos0:
prevents having to use `[Title|Description](fmt.Sprinf(...))`.


maaslalani:
This is really convenient but I would rather keep the surface area of the API as small as possible so it's easier to find stuff and not accidentally autocomplete things.

That being said we might need this for the dynamic updating stuff so it's possible we'll have to re-open this.

PR #102, opened 2024-01-16
fix incorrect documentation for the `WithShowErrors` method

theredditbandit:
currently the documentation for the [`Form.WithShowErrors()` ](https://github.com/charmbracelet/huh/blob/58163e7b5b2ffb7d60176490eda32ee4210b9d40/form.go#L170) is the same as [`Form.WithShowHelp`](https://github.com/charmbracelet/huh/blob/58163e7b5b2ffb7d60176490eda32ee4210b9d40/form.go#L181) , this fixes that.

### DIFF
--- a/form.go
+++ b/form.go
@@ -176,7 +176,7 @@ func (f *Form) WithShowHelp(v bool) *Form {
 
 // WithShowErrors sets whether or not the form should show errors.
 //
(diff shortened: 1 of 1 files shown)


maaslalani:
Thanks so much @theredditbandit! We really appreciate the contribution!

PR #103, opened 2024-01-17
fix: shift+tab would fail if first group is hidden

caarlos0:
the previous version was changing the page to a potentially hidden page, this PR fixes it, and covers it with tests.


maaslalani (approved):
Thank you for adding these tests!

PR #104, opened 2024-01-17
Next / Previous dynamic help

maaslalani:
Change keymap depending on field position.

### DIFF
--- a/form.go
+++ b/form.go
@@ -131,7 +131,7 @@ type Field interface {
 	WithHeight(int) Field
 
 	// WithPosition tells the field the index of the group and position it is in.
-	WithPosition(Position) Field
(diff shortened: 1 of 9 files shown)


form.go
```diff
@@ -129,13 +130,26 @@ type Field interface {
...
+
 	// GetKey returns the field's key.
 	GetKey() string
 
 	// GetValue returns the field's value.
 	GetValue() any
 }
 
+// Position is positional information about the given field and form.
+type Position struct {
```
caarlos0, line 144: since all fields are private, does this struct, and `WithPosition`, need to be public? 🤔

maaslalani: Exposed all the fields since this is an interface and could be used by others I think we need to make these public.

PR #109, opened 2024-01-19
feat: allow to pass custom tea.ProgramOptions

### LINKED ISSUE #101: Set Tea program options
It would be nice to be able to set Tea program options. Currently we don't have to way to do that since programs are instantiated during `run()`.

One way would be to provide a `RunWith(*tea.Program)` method or something equivalent.

My use case is the following:
I'm using Huh for a CLI application and when running under CI I don't have access to a TTY. I'd like to be able to run my tests using the `WithInput` option to prevent the program from attempting to open a TTY.

Thanks

caarlos0:
Possible usecases:

- custom input/output
- disable renderer
- ...

closes #101

### DIFF
--- a/examples/custom-tea-opts/main.go
+++ b/examples/custom-tea-opts/main.go
@@ -1,0 +1,19 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/charmbracelet/huh"
+)
+
+func main() {
+	var name string
+	if err := huh.NewForm(
+		huh.NewGroup(
+			huh.NewInput().Description("What is your name?").Value(&name),
+		),
+	).Run(); err != nil {
+		fmt.Println("error:", err)
+	}
+	fmt.Println("Hello, " + name + "!")
+}
--- a/field_confirm.go
+++ b/field_confirm.go
@@ -176,11 +176,11 @@ func (c *Confirm) View() string {
 }
 
 // Run runs the confirm field in accessible mode.
-func (c *Confirm) Run() error {
+func (c *Confirm) Run(opts ...tea.ProgramOption) error {
 	if c.accessible {
 		return c.runAccessible()
 	}
-	return Run(c)
+	return Run(c, opts...)
 }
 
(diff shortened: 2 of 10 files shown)


examples/custom-tea-opts/main.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/charmbracelet/huh"
+)
+
+func main() {
+	var name string
+	if err := huh.NewForm(
+		huh.NewGroup(
+			huh.NewInput().Description("What is your name?").Value(&name),
+		),
+	).Run(); err != nil {
```
meowgorithm, line 15: This example is missing the actual options, right?

meowgorithm:
Well this is awesome. My only thought here would something like `.WithProgramOptions()` be more idiomatic for this library?

examples/custom-tea-opts/main.go
caarlos0: hah, lol, yeah, was testing different ones and commited without any, fixing

caarlos0:
> Well this is awesome. My only thought here would something like `.WithProgramOptions()` be more idiomatic for this library?

hmmm, I dunno... it kinda makes sense to have it on `Run` IMHO... @maaslalani any opinions on this?

jon4hz:
Hey, what's the status on this? I'd love to have an option to display `huh` on stderr, so you can use it in programs that redirect stdout. :)

maaslalani:
> Well this is awesome. My only thought here is would something like `.WithProgramOptions()` be more idiomatic for this library?

Yep, I totally agree with this, will update!

maaslalani:
Fixed by: https://github.com/charmbracelet/huh/commit/ccc1284bb3b60eb3bf0b4e3f21f9eac5939ce226

I made it so that it's not applicable on the short form of `Run`. If the user wants to specify the `tea.ProgramOption`s they'll need to create a full form. Might change this later.

jon4hz:
Amazing! Thanks so much for the quick fix! ^^

PR #117, opened 2024-01-23
group: do not render gap for empty help view

ardnew:
`(*Group).View` renders enabled keybindings in a footer view according to the current `(Field).KeyBinds()`. This footer view is offset vertically by at least one unconditional `(Theme).FieldSeparator`. 

This is fine in most cases as the user will usually have key bindings returned by `(Field).KeyBinds()` for every `Field` – and indeed every package-provided `Field` returns something meaningful when they do.

However, if the user wishes to **not** render this footer view in a custom `Field` (by returning a nil or empty slice from `(Field).KeyBinds()` and using the `KeyMap` directly in `Update`), then the renderer can get confused about the intended output height.

This PR adds a condition on rendering the footer view and its `(Theme).FieldSeparator` gap to ensure they contain more than just white space when combined together.


ardnew:
Yikes, rebased against the wrong remote. Need to rebase again to squash all these changes.

maaslalani:
Hey @ardnew, let me know if you're able to rebase / open another PR for this one!

ardnew:
Man, I'm having a lot of trouble fixing this headache. I may close this PR and open another one with a much cleaner patch. Is this okay with you? @maaslalani

maaslalani:
Of course! @ardnew

ardnew:
Oh, just re-read your previous comment suggesting I open another PR as an option. I'm gonna take that route. :wink:

PR #118, opened 2024-01-24
multiselect: set height in WithHeight

ardnew:
This one should be pretty clear :wink:


maaslalani:
Nice catch, thank you!

PR #122, opened 2024-01-24
multiselect: set height in WithHeight

ardnew:
Amend to the previous PR, which forgot to also update the godoc comment from "width" to "height"


maaslalani:
Thanks so much @ardnew!

Issue #73, opened 2023-12-18
Cannot override `Text` `KeyMap`


guigui64:
**Describe the bug**
`WithKeyMap` does not seem to have any effect on the `Text` field.

**To Reproduce**
```go
textField := huh.NewText().
	Title("...").
	Value(&message).
	WithKeyMap(&huh.KeyMap{
		Text: huh.TextKeyMap{
			Next:    key.NewBinding(key.WithKeys("ctrl+enter"), key.WithHelp("ctrl+enter", "next")),
			Prev:    key.NewBinding(key.WithKeys("shift+tab"), key.WithHelp("shift+tab", "back")),
			NewLine: key.NewBinding(key.WithKeys("enter", "ctrl+j"), key.WithHelp("enter / ctrl+j", "new line")),
			Editor:  key.NewBinding(key.WithKeys("ctrl+e"), key.WithHelp("ctrl+e", "open editor")),
		},
	})

form := huh.NewForm(
	huh.NewGroup(
		textField,
	),
)
```

Help text at the end of the field does not reflect the new KeyMap and the default keymap still applies.

**Expected behavior**
Next and NewLine should have been overriden.

meowgorithm:
Hi! Thanks for the report. For now, you can change the keymap on the form to achieve the same effect:

```go
textField := huh.NewText().
	Title("...").
	Value(&message)

form := huh.NewForm(
	huh.NewGroup(
		textField,
	),
).WithKeyMap(&huh.KeyMap{
	Text: huh.TextKeyMap{
		Next:    key.NewBinding(key.WithKeys("ctrl+enter"), key.WithHelp("ctrl+enter", "next")),
		Prev:    key.NewBinding(key.WithKeys("shift+tab"), key.WithHelp("shift+tab", "back")),
		NewLine: key.NewBinding(key.WithKeys("enter", "ctrl+j"), key.WithHelp("enter / ctrl+j", "new line")),
		Editor:  key.NewBinding(key.WithKeys("ctrl+e"), key.WithHelp("ctrl+e", "open editor")),
	},
})
```

caarlos0:
closing per @meowgorithm response

PR #127, opened 2024-01-26
delegate key bindings shown in Field's help views

ardnew:
The goal of this PR is to allow `Field` implementations to specify which key bindings appear in both the full-length help view as well as the short, one-line help view — which would enable: 

1. Context-driven help
2. Mock keybindings (e.g., `key.NewBinding(key.WithHelp("←↑↓→", "move"))`)
3. Managing visibility of disabled, duplicate, non-essential, or unpublished key bindings
4. Delegating logic to lower-layer `Field` to reduce complexity of higher-layer (`Group`, `Form`, `Bubble`, etc.)

This change adds a new method to the `Field` interface:

```go
	// KeyBindsHelp returns the help keybindings.
	KeyBindsHelp(HelpFormat) []key.Binding
```

Along with a new exported type used as parameter to the method above:

```go
// HelpFormat is the way in which help is presented.
type HelpFormat int

const (
	// FormatLong is the full-size default help format.
	FormatLong HelpFormat = iota

	// FormatShort is the one-line summary help format.
	FormatShort
)
```

The existing method to access key bindings remains unmodified for backwards-compatibility. Also, this method is still useful and meaningful as the correct interface to retrieve _all defined_ key bindings from a `Field`:

```go
	// KeyBinds returns all keybindings.
	KeyBinds() []key.Binding
```

However, this PR does **not** change any implemented `Field`'s current behavior or help views. It is only concerned with implementing the functionality for them to be _able_ to define their own help views. Thus:

 - No existing tests need to be modified
 - Each of the package-provided `(Field).KeyBindsHelp` methods are implemented similar to: 
   -    ```go
        // KeyBindsHelp returns the help keybindings for the text field.
        func (t *Text) KeyBindsHelp(HelpFormat) []key.Binding {
        	return t.KeyBinds() // ... return appropriate keys based on state of receiver
        }
        ```


ardnew:
An alternative to `(Field).KeyBindsHelp` I considered is to instead provide `(Field).SetKeyBindsCallback(func() []key.Binding)`, thus allowing the user to define `(Field).KeyBinds` in any way they choose. 

This would enable custom logic to produce keybindings, and it would allow overriding package-defined keybindings (which would fix #73).

However, apart from being generally more complex and error-prone, the principal drawback to this approach is its inability to distinguish if keybindings are intended for key handling versus help content (and for which help view: short or long?)

maaslalani:
Hey @ardnew, let me know what you think but now we have dynamic help options which change based on the fields position and make everything feel more responsive. This complicates a lot of the logic internally though since we're always changing and checking the field position to update the help.

ardnew:
Closing this PR due to a difficult-to-fix rebase mistake. 

Opened new issue #202 to address the @maaslalani comment above. Will open a new PR if necessary following issue discussion.

PR #130, opened 2024-01-31
feat: Add a way to completely hide password from Input

### LINKED ISSUE #129: Support textinput.EchoNone in huh.NewInput().Password()
Hi, thanks for that lib, which seems to contain so interesting stuff. 
We'd like still to have a very little feature that would clearly be an enabler for us.

**Is your feature request related to a problem? Please describe.**
We don´t find any good lib rendering correctly the password. Every time, it is proposed to replace characters by a mark. This is not what we'd like but something more like the ``EchoNone`` proposed by the bubble library. Why not propose it as well in huh?

**Describe the solution you'd like**
I'd like a setter similar with Password() one but that would set ``EchoNone`` as ``EchoMode`` under the hood.

**Describe alternatives you've considered**
I tried to create a custom InputField without success.
I finally ending up using huh for all my fields except password for which I use raw bubble directly.

Anyway simple setter, like Password() is doing, would be much more readable.

**Additional context**
No further context

**Edit**:
To be 100% clear, I'm speaking about these lines of code https://github.com/charmbracelet/huh/blob/main/field_input.go#L103-L110

celian-garcia:
fixes #129

### DIFF
--- a/field_input.go
+++ b/field_input.go
@@ -109,6 +109,17 @@ func (i *Input) Password(password bool) *Input {
 	return i
 }
 
+// ExtraHiddenPassword sets whether or not to hide the input while the user is typing.
+// Like Password but in this case without displaying mark on each character.
+func (i *Input) ExtraHiddenPassword(password bool) *Input {
+	if password {
+		i.textinput.EchoMode = textinput.EchoNone
+	} else {
+		i.textinput.EchoMode = textinput.EchoNormal
+	}
+	return i
+}
+
 // Placeholder sets the placeholder of the text input.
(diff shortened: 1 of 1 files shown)


celian-garcia:
> Hey @celian-garcia, thank you so much for this PR. #131 should solve your issues, we decided to go with this to be more consistent with the [Bubbles](https://github.com/charmbracelet/bubbles) API.

That's perfect for me, even better :) thank you so much guys for your reactivity

maaslalani:
Thanks again for the PR and for using `huh`! Let us know if you have any more feedback.

PR #136, opened 2024-02-01
Inline Select

maaslalani:
Allows `Select` field to be placed `inline`.


maaslalani:
image

theme.go
```diff
@@ -48,6 +48,8 @@ type FieldStyles struct {
 	// Select styles.
 	SelectSelector lipgloss.Style // Selection indicator
 	Option         lipgloss.Style // Select options
+	SelectNext     lipgloss.Style
+	SelectPrev     lipgloss.Style
```
maaslalani, line 52: @meowgorithm, not 100% sure on these names. Do you have any thoughts?

Maybe something more generic such as `NextIndicator` `BackIndicator`?

meowgorithm: Good call. How about `NextIndicator` and `PrevIndicator`?

maaslalani: Yup, sounds good to me!

Issue #134, opened 2024-01-31
Predefined Validation Functions


maaslalani:
Huh should define some common validation so that users can use them easily without having to write their own.

For example:

* `huh.ValidateNotEmpty`
* `huh.ValidateLength(min, max int)`
* `huh.ValidateOneOf(options ...string)`
* etc...

```
huh.NewInput().Title("What's your name?").Validate(huh.ValidateLength(3, 50)
```

maaslalani:
Fixed by #140, thanks so much again @anirudhaCodes! ❤️

PR #140, opened 2024-02-02
predefined Validation Functions

anirudhaCodes:
This PR refers to issue #134 .

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -5,11 +5,13 @@ import (
 	"unicode/utf8"
 )
 
+var ErrInputEmpty = fmt.Errorf("input cannot be empty")
+
 // ValidateNotEmpty checks if the input is not empty.
 func ValidateNotEmpty() func(s string) error {
 	return func(s string) error {
 		if s == "" {
-			return fmt.Errorf("input cannot be empty")
+			return ErrInputEmpty
 		}
 		return nil
 	}
@@ -35,7 +37,7 @@ func ValidateOneOf(options ...string) func(string) error {
 
 	return func(value string) error {
 		if _, ok := validOptions[value]; !ok {
-			return fmt.Errorf("invalid option")
+			return fmt.Errorf("invalid option: %s", value)
 		}
 		return nil
 	}
--- a/validate.go
+++ b/validate.go
@@ -2,20 +2,23 @@ package huh
 
 import (
 	"fmt"
+	"unicode/utf8"
 )
 
 // ValidateNotEmpty checks if the input is not empty.
-func ValidateNotEmpty(s string) error {
-	if s == "" {
-		return fmt.Errorf("input cannot be empty")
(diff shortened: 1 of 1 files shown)


validate.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+	"unicode/utf8"
+)
+
+// ValidateNotEmpty checks if the input is not empty.
+func ValidateNotEmpty() func(s string) error {
+	return func(s string) error {
+		if s == "" {
+			return fmt.Errorf("input cannot be empty")
```
caarlos0, line 12: I think it would be nice to export these errors, so users can use `errors.Is` with them 🤔

validate.go
```diff
@@ -0,0 +1,42 @@
...
+// ValidateOneOf checks if a string is one of the specified options.
+func ValidateOneOf(options ...string) func(string) error {
+	validOptions := make(map[string]struct{})
+	for _, option := range options {
+		validOptions[option] = struct{}{}
+	}
+
+	return func(value string) error {
+		if _, ok := validOptions[value]; !ok {
+			return fmt.Errorf("invalid option")
```
caarlos0, line 38:
```suggestion
			return fmt.Errorf("invalid option: %s", value)
```

validate.go
anirudhaCodes: Yes make sense.

I was also thinking that error message can be a input. Because the error message might depend on final use case.

maaslalani:
Hey @anirudhaCodes thanks for this PR.

I haven't fully decided on the naming of these functions so they might change.

One thing so would suggest is to defined a `ValidateMinLength` function and a `ValidateMaxLength` function and then use these two functions to create a `ValidateLength` function and also Alias `ValidateNotEmpty` to `MinLength(1)`, etc...

That way we can make all the errors consistent and nicer and the functions easier to test.

validate.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// ErrInputEmpty is an error indicating that the input cannot be empty.
+var ErrInputEmpty = fmt.Errorf("input cannot be empty")
+
+// ValidateNotEmpty checks if the input is not empty.
+func ValidateNotEmpty() func(s string) error {
+	return func(s string) error {
+		if err := ValidateMinLength(1)(s); err != nil {
+			return ErrInputEmpty
```
maaslalani, line 15:
```suggestion
			return fmt.Errorf("input cannot be empty")
```

validate.go
```diff
@@ -0,0 +1,64 @@
+package huh
+
+import (
+	"fmt"
+	"unicode/utf8"
+)
+
+// ErrInputEmpty is an error indicating that the input cannot be empty.
+var ErrInputEmpty = fmt.Errorf("input cannot be empty")
+
```
maaslalani, line 10:
```suggestion
```

PR #141, opened 2024-02-06
feat: allow to use a custom lipgloss.Renderer

caarlos0:
this would allow to set a custom lipgloss.Renderer into a theme, which is particularly useful if running the form in a wish app.

PS: this won't work properly because the bubbles don't support it yet. We'll discuss a way forward tomorrow cc/ @muesli @meowgorithm


maaslalani:
Closing for now since I think we're going to be doing this at the Bubble Tea / Bubbles level? If not please re-open.

PR #143, opened 2024-02-09
feat: ctrl+u, ctrl+d, g, and G keybindings on selects

caarlos0:
i use vim, btw

Not sure if we need/want to show them on the help, as the help would then use a lot of space...

If you agree we shouldn't, I'll remove the commented stuff.

### DIFF
--- a/field_multiselect.go
+++ b/field_multiselect.go
@@ -168,7 +168,21 @@ func (m *MultiSelect[T]) Blur() tea.Cmd {
 
 // KeyBinds returns the help message for the multi-select field.
 func (m *MultiSelect[T]) KeyBinds() []key.Binding {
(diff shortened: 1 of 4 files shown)


maaslalani:
Nice! Thank you, I was missing these as well. Agreed we shouldn't crowd the help, we can keep these hidden.

field_multiselect.go
```diff
@@ -168,7 +168,21 @@ func (m *MultiSelect[T]) Blur() tea.Cmd {
...
 func (m *MultiSelect[T]) KeyBinds() []key.Binding {
-	return []key.Binding{m.keymap.Toggle, m.keymap.Up, m.keymap.Down, m.keymap.Filter, m.keymap.SetFilter, m.keymap.ClearFilter, m.keymap.Prev, m.keymap.Submit, m.keymap.Next}
+	return []key.Binding{
+		m.keymap.Toggle,
+		m.keymap.Up,
+		m.keymap.Down,
+		// m.keymap.HalfPageUp,
+		// m.keymap.HalfPageDown,
+		// m.keymap.GotoBottom,
+		// m.keymap.GotoTop,
```
maaslalani, line 178:
```suggestion
```

Issue #69, opened 2023-12-18
Disply help message when not using form


tetienne:
**Is your feature request related to a problem? Please describe.**
My current CLI ask several questions to the end user, and each question depends on the choice of the previous one. For instance, select an AWS cluster, then an AWS service.
So I cannot use the form feature of huh (or can I?). Because of this, I cannot display the help message for my prompt.

**Describe the solution you'd like**
Add a parameter ShowHelpMessage for the fields also.

**Describe alternatives you've considered**
Create a form for each single question.

**Additional context**
N/A

meowgorithm:
Hi! This is something we'd to explore in a more official capacity, but for now you can achieve a similar effect using multiple forms. For an example to help get started see https://github.com/charmbracelet/huh/pull/75.

tetienne:
Thx for the exemple. So no magic yet here :)

alinagard:
+1 to this enhancement!!

kpym:
You can may be use `WithHideFunc` to conditionally show some questions. Here is the rewriten [example](https://github.com/charmbracelet/huh/blob/0f20f61417bfee7a1ec6f0f2f93a423d884b20ad/examples/conditional/main.go) of #75 (with corrected "Carrot" option for "vegetables" 🙂).

```golang
package main

import (
	"fmt"

	"github.com/charmbracelet/huh"
)

func main() {

	var (
		category string
		choice   string
	)

	huh.NewForm(
		huh.NewGroup(
			huh.NewSelect[string]().
				Title("What are you in the mood for?").
				Value(&category).
				Options(
					huh.NewOption("Some fruit", "fruits"),
					huh.NewOption("A vegetable", "vegetables"),
					huh.NewOption("A drink", "drinks"),
				),
		),
		huh.NewGroup(
			huh.NewSelect[string]().
				Title("Okay, what kind of fruits are you in the mood for?").
				Value(&choice).
				Options(
					huh.NewOption("Tangerine", "tangerine"),
					huh.NewOption("Canteloupe", "canteloupe"),
					huh.NewOption("Pomelo", "pomelo"),
					huh.NewOption("Grapefruit", "grapefruit"),
				),
		).WithHideFunc(func() bool {
			return category != "fruits"
		}),
		huh.NewGroup(
			huh.NewSelect[string]().
				Title("Okay, what kind of vegetables are you in the mood for?").
				Value(&choice).
				Options(
					huh.NewOption("Carrot", "carrot"),
[...]

PR #144, opened 2024-02-09
feat: select onchange

caarlos0:
The idea is to allow one select to depend on another, e.g., filtering country, then states within that country, then cities within that state.

This is just a raw idea to get your opinions on it, so no tests or anything yet.

### DIFF
--- a/field_select.go
+++ b/field_select.go
@@ -95,6 +95,13 @@ func (s *Select[T]) Description(description string) *Select[T] {
 	return s
 }
 
+// SetOptions allows to change the available options of a select at runtime.
+// This will also reset the selected item.
+func (s *Select[T]) SetOptions(options ...Option[T]) {
+	s.Options(options...)
+	s.selected = 0
+}
+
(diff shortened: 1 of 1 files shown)


maaslalani:
Personally, I think we should go for a more declarative approach.

Perhaps allowing the options to be passed an `OptionsFunc` (and maybe allowing it to specify dependencies, i.e. which values change).

maaslalani:
So for the `Country / State` example maybe it would be something like:

```
var country string

huh.NewSelect[string]().Title("Country").Options(...)

huh.NewSelect[string]().Title("State").OptionsFunc(..., &country)
```

caarlos0:
That wouldn't be the same though, as `country` would only change when the select is confirmed, not while "browsing", e.g.:

CleanShot 2024-02-12 at 10 43 20

maaslalani:
Taking the `OptionsFunc` declarative route here:
This will support API calls and handle caching for the user as well.

* #69

PR #148, opened 2024-02-13
feat: add GetFocused to field_input

SpyMachine:
My use case is that I have a field that I'd like to update suggestions for based on what the user has input. When the user starts typing, a search endpoint is hit with the current value of the field as the query parameter and the suggestions is updated based off the returned search.

So basically, in my Update I want to check if the field is focused and then if a new key was pressed to hit the endpoint and update the suggestions.

### DIFF
--- a/field_input.go
+++ b/field_input.go
@@ -342,3 +342,8 @@ func (i *Input) GetKey() string {
 func (i *Input) GetValue() any {
 	return *i.value
 }
+
+// GetFocused returns if the input is focused.
+func (i *Input) GetFocused() bool {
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey @SpyMachine, dynamic inputs are now merged. Let me know if this example is what you were trying to accomplish:

https://github.com/charmbracelet/huh/blob/main/examples/dynamic/dynamic-suggestions/main.go

caarlos0:
Closing as per last comment.

Let me know if this is still an issue.

Thanks!

Issue #149, opened 2024-02-18
Filter options inside a select field based in the previous steps data


lucasfloriani:
**Is your feature request related to a problem? Please describe.**
I'm implementing a project where it's a CLI that will create shortcuts with the keyboard that will run complex tasks. There's one flow where we can select the events we want and which keys will trigger them in the same form process. We currently have 5 options available where you can select the key to press, but we want to make it more user-friendly and filter the keys that the customer has already selected to execute something.

**Describe the solution you'd like**
As we are familiar with the WithHideFunc at the group level, I believe it would be nice to have something like that at the Option level.

**Describe alternatives you've considered**
I was thinking of using WithHideFunc at the group level to create a brute force of all possibilities, but it will create a lot of boilerplate for a simple functionality

PS: If we have another way to implement such a feature with the actual API please let me know 🙇‍♂️ 

Also, thanks for all those amazing libraries for CLI. It made me go back to Golang to have some fun 😄

maaslalani:
Hey @lucasfloriani, this is a great idea. Thanks for opening up this issue. I'm going to merge this issue with a similar one to consolidate the ideas:

* https://github.com/charmbracelet/huh/issues/69

The system we want to create is a way to dynamically change all fields by reacting to previous answers (or perhaps even other events, timeouts? external APIs? etc...) which will allow for very feature-rich and functional forms.

Feel free to re-open this issue or make a new one if you don't think your use case will be solved.

PR #151, opened 2024-02-19
[Feature] Add hide support on Options

lucasfloriani:
This code will implement the functionality I described in the issue #149

### DIFF
--- a/examples/conditionalOption/main.go
+++ b/examples/conditionalOption/main.go
@@ -1,0 +1,275 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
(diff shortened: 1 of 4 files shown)


maaslalani:
Hey @lucasfloriani, I'm not 100% sure if we'll go with this approach, but do you mind changing this example to use Countries and States / Provinces (sampled) so that it's more clear how this all works.

maaslalani:
Hey @lucasfloriani, thanks so much for your work on this:

image


We're going to go with an approach with `OptionsFunc` which allows things to be more dynamic and (IMO) a little more clearer and idiomatic. Thanks again for your idea and implementation!

PR #154, opened 2024-02-26
fix: remove temp file after opening external editor in `Text` field

auvred:


maaslalani:
Thanks so much @auvred! ❤️

PR #156, opened 2024-02-29
Themes should be applied as specified

maaslalani:
This applies themes as specified, in order of specificity.

Themes will be applied to Fields, then Groups, then Forms. This way different fields / groups can have different themes.


maaslalani:
cc @jon4hz

This should fix all the theming issues, thank you so much for catching this!

maaslalani:
Actually, it's a bit more complicated than specified here. Will have to rework things a little bit. Mainly we need a distinction between default themes / theme not being set.

PR #172, opened 2024-03-19
Added optional printer argument to spinner action

calebstewart:
This PR adds an optional argument to the printer action which provides access to the `tea.Printf` and `tea.Println` functionality. I added the `Printer` interface with two methods: `Printf` and `Println`. I also added a `accessiblePrinter` type which just proxies the calls to `fmt.Printf` and `fmt.Println`.

When called normally (without accessibility), the `tea.Program` is passed in as the Printer object. When called with `Accessible(true)`, then a dummy `accessiblePrinter` object is passed in.

The Spinner interface is left backwards compatible. The `Action()` method still takes a `func()`, but a new method is introduced called `ActionWithPrinter()` which takes a `func(Printer)` function instead.


maaslalani:
Ohhh, this is a very nice idea and good implementation @calebstewart! Thank you so much.

Let me just think about what we should call this to make it more easily discoverable. I almost wonder if we could have people simply print to a buffer on the spinner to achieve this effect i.e. with `fmt.Sprintln(spinner.Out, "This is a test")` 🤔

caarlos0:
this should have been fixed in #292 

thanks for the PR 🙏🏻

PR #195, opened 2024-04-21
feat: disabled options

### LINKED ISSUE #77: Add Disabled State to Options
**Is your feature request related to a problem? Please describe.**
We want to provide multiselects to users where some of the options are disabled.
Generally speaking we can filter our disabled options but for our use-case it would feel like missing context for our users.
It would be better if they could see the state of all their options even if specific options have been disabled from selection.

**Describe the solution you'd like**
It would be great if we could do:
```go
 huh.NewMultiSelect[string]().
    Title("Toppings").
    Options(
        huh.NewOption("Lettuce", "lettuce").Selected(true),
        huh.NewOption("Tomatoes", "tomatoes").Disabled(true),
        ...
```

The result would be that Tomatoes may appear gray and/or strikethrough and could not be selected.
Perhaps the arrow functions would skip over its index entirely, or selection would be blocked or trigger some visual element.
I have no hard requirements as to user experience.

gabrielchl:
summary:
- add disabled options (used in select and multi-select components)
- these options will be rendered (including the filtered list), but un-toggle-able
  - select: you can't select that option
  - multi-select: you can't toggle that option 
- color-wise, they use the same color description does

(i'm not sure how the gif is best generated for the readme)

example usage:
```go
huh.NewOption("Charm Sauce (sorry, we've run out!)", "Charm Sauce").Disabled(true),
```

example form question:
```
Toppings:
> [ ] Lettuce
  [ ] Tomatoes
  [ ] Charm Sauce (sorry, we've run out!)
  [ ] Nutella
```

example GIF:
Untitled

closes #77

### DIFF
--- a/field_multiselect.go
+++ b/field_multiselect.go
@@ -255,7 +255,8 @@ func (m *MultiSelect[T]) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		case key.Matches(msg, m.keymap.Toggle):
 			for i, option := range m.options {
 				if option.Key == m.filteredOptions[m.cursor].Key {
-					if !m.options[m.cursor].selected && m.limit > 0 && m.numSelected() >= m.limit {
+					if !m.options[m.cursor].selected && m.limit > 0 && m.numSelected() >= m.limit ||
+						m.options[m.cursor].disabled {
 						break
 					}
 					selected := m.options[i].selected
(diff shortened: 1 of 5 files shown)


maaslalani:
Hi, thank you so much for the PR. This is an awesome idea and we might revisit this in the future but for now I'd like to avoid allowing disabled options in forms simply out of personal preference. Dynamic huh now allows you to remove the options you aren't able to select and re-add then when appropriate.

PR #200, opened 2024-04-29
fix(group): do not render newline before empty help

ardnew:
This is the same change as PR #117, which was closed due to my incompetence with git. This is now baselined against the latest `main` (f922e26dffc13c3bcea280c40865c110611cf257).

Note that since #117, upstream `charmbracelet/huh` changed the separator "gap" between group content view and keybindings/errors view from using the current theme (`g.theme.FieldSeparator.String()`) to always using the literal line feed (`'\n'`). Assuming that was intentional, I've maintained that behavior in this PR.

### DIFF
--- a/group.go
+++ b/group.go
@@ -2,6 +2,7 @@ package huh
 
 import (
 	"strings"
+	"unicode"
 
 	"github.com/charmbracelet/bubbles/help"
 	"github.com/charmbracelet/bubbles/paginator"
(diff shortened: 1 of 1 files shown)


group.go
```diff
@@ -303,17 +304,26 @@ func (g *Group) buildView() {
...
-		view.WriteString(g.help.ShortHelpView(g.fields[g.paginator.Page].KeyBinds()))
+		foot.WriteString(g.help.ShortHelpView(g.fields[g.paginator.Page].KeyBinds()))
 	}
 	if g.showErrors {
 		for _, err := range errors {
-			view.WriteString(g.theme.Focused.ErrorMessage.Render(err.Error()))
+			foot.WriteString(g.theme.Focused.ErrorMessage.Render(err.Error()))
 		}
 	}
+	if foot.Len() > 0 && hasNonSpace(foot.String()) {
```
maaslalani, line 320: I think you could also simply check:
```suggestion
	if strings.TrimSpace(foot.String()) != "" {
```

maaslalani:
Do you mind posting an example of a before and after screenshot of which scenario view this fixes, thank you so much!

group.go
ardnew: Errr, right, of course. Thanks 😇

`hasNonSpace` seems a little silly at second glance. Will remove its definition entirely

PR #215, opened 2024-05-03
feat: custom renderer in theme

adamdottv:
i should state up front that i'm not sure if this is a desirable change, and i considered opening an issue instead but thought a pr might be easier to start the conversation with. no worries if you'd rather not introduce this change.

we're leveraging `wish` and `bubbletea` over @terminaldotshop and i noticed that our `huh` form colors were slightly off from the rest of our app, despite having configured them with the same `lipgloss.Color` values. i ended up copying over a bunch of the theme stuff into our repo in order to use the same `lipgloss.Renderer` and it solved the issue, so i'm bubbling up that change here. i wanted to avoid introducing a breaking change, so went with this approach, but it also looks a little clunky. lmk what you think!

### DIFF
--- a/theme.go
+++ b/theme.go
@@ -126,42 +126,42 @@ const (
 	buttonPaddingVertical   = 0
 )
 
-// ThemeBase returns a new base theme with general styles to be inherited by
-// other themes.
-func ThemeBase() *Theme {
+// ThemeBaseWithRenderer returns a new base theme with general styles to be inherited by
+// other themes, with a custom lipgloss renderer.
+func ThemeBaseWithRenderer(lg *lipgloss.Renderer) *Theme {
 	var t Theme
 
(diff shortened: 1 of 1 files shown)


caarlos0:
refs https://github.com/charmbracelet/huh/pull/141

maaslalani:
Thanks so much for this change! I will add an SSH example to `huh` to test this out.

maaslalani:
Simple SSH form example:

https://github.com/charmbracelet/huh/pull/216

adamdottv:
#141 looks like a better (and more conplete) implementation! did the reason for closing it still apply?

maaslalani:
Yeah I think we should either go with #141 or #217! I think the team is leaning towards #141

adamdottv:
went ahead and closed this, i'll track the other PRs to see what lands!

PR #222, opened 2024-05-07
Add dynamic content option to note input

abtmr:
Hi, thanks for huh!

I would like the ability to show content dynamically within a form which I think could fit nicely into the note field. I've added an option to pass a `func() string` to a `WithContent` func, which is used to fetch the dynamic content. The example provided adds a note to the burger example, showing the selection before completing the order.

I also added a `WithNextLabel` func so we can choose the next button label too.

Thanks again!

### DIFF
--- a/examples/burger/main.go
+++ b/examples/burger/main.go
@@ -118,6 +118,26 @@ func main() {
 				Description("You get one free side with this order."),
 		),
 
+		// Confirm the customer is happy with their selection so far.
+		huh.NewGroup(
+			huh.NewNote().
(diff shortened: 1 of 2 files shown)


maaslalani:
Hey @abtmr, thanks so much for this PR. We're actually implementing a full system to make anything dynamic in #69, this will include Note titles, descriptions, etc...

abtmr:
thanks for the reply, dynamic branch looks great (wish I'd checked what was going on in branches first)! ok, I'll close this.

PR #225, opened 2024-05-08
feat: optional setting of note next label

abtmr:
I'd like to be able to set the label on the note next button. This change allows this through a `WithNextLabel` func. If this isn't used, then will default to 'Next'.

### DIFF
--- a/examples/burger/main.go
+++ b/examples/burger/main.go
@@ -59,7 +59,10 @@ func main() {
 	form := huh.NewForm(
 		huh.NewGroup(huh.NewNote().
 			Title("Charmburger").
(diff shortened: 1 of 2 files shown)


maaslalani:
This is awesome! Thank you, @abtmr! I would probably call the function `NextLabel` and return the `*Note` instead of the `Field` so people can stack the properties correctly.

abtmr:
Ah yeah good spot, that's a copy/paste mistake... I'll rename and change it to return the note! Thanks

maaslalani:
Awesome work on this one @abtmr!

PR #234, opened 2024-05-11
feat: expose results by shallow copy

hay-kot:
I am migrating my scaffolding tool to use `huh` and am running into a few limitations. One of which is that, that I don't know what my form will look likes keys or values at build time, only at runtime do I know. The Fields are generated dynamically by a user configuration file. 

As such, I need some way to get that data out of the form to continue onto the template engine. 

This PR exposes the underlying map inside of the `Form` type by shallow copy. I think in most cases, this should be fine?

--

I can work around this limitation by saving a copy of the keys and getting them out of the `Form` using the existing `GetValue` method, but this saves me a few steps. If you want to keep this internal only let me know! 

Love the project!

### DIFF
--- a/form.go
+++ b/form.go
@@ -586,3 +586,14 @@ func (f *Form) runAccessible() error {
 
 	return nil
 }
+
+// Results returns a shallow copy of the results map.
+func (f *Form) Results() map[string]any {
+	cp := make(map[string]any, len(f.results))
+
+	for k, v := range f.results {
+		cp[k] = v
+	}
+
+	return cp
+}


meowgorithm:
Hi! Thanks for this one. So we can understand it better, mind sharing some code to help illustrate your workflow and how this solves it (we try and be fairly conservative about increasing API surface area).

hay-kot:
Yeah! I actually was able to work around the limitation I was running into so I don't think this would really help me much anymore. I am doing _gross_ things to accomplish my goal, so I'll share my use-case and what I did, at the very least maybe it will help someone in the future! 

**Goal**

User's use YAML to define customizable forms for collecting input. Each field as the possibility of having a condition go template (e.g. `{{ .render_next_question }}` that get's evaluated before being rendered, this template requires the variables from the previous question(s). 

**Solution**

I was able to accomplish by wrapping questions in a custom type, and using a hook to assign the fields value to a map that I manage. Then I call those hooks in the `WithHideFunc` associated with the group. 

This get's the job done, but I don't _love_ it. 

- [Form Logic](https://github.com/hay-kot/scaffold/blob/705389d34e7b94cd12177ced407f334b705e7970/app/scaffold/project.go#L145-L191)
- [Custom Wrapper](https://github.com/hay-kot/scaffold/blob/main/app/scaffold/askable.go#L11)
- [Example Hook](https://github.com/hay-kot/scaffold/blob/705389d34e7b94cd12177ced407f334b705e7970/app/scaffold/questions.go#L174-L177)

PR #237, opened 2024-05-15
Spinner bubbles up `tea.Program` errors

clowder:
When running the tea.Program backing the spinner generates an error then bubble it up to the user.

Got bitten by this when a user of our CLI tried running it via `ssh`. The underlying `tea.Program` was throwing an error about not having a TTY, but the CLI was just silently eating the error.

### DIFF
--- a/spinner/spinner.go
+++ b/spinner/spinner.go
@@ -151,9 +151,9 @@ func (s *Spinner) Run() error {
 		}()
 	}
 
-	_, _ = p.Run()
+	_, err = p.Run()
 
-	return nil
+	return err
 }
 
 // runAccessible runs the spinner in an accessible mode (statically).


clowder:
In our case we're working around it for now by running the program directly.

```go
s := spinner.New().Title("Waiting…")
p := tea.NewProgram(s, tea.WithContext(ctx), tea.WithOutput(os.Stderr))
_, err := p.Run()
```

spinner/spinner.go
```diff
@@ -151,9 +151,9 @@ func (s *Spinner) Run() error {
 		}()
 	}
 
-	_, _ = p.Run()
+	_, err = p.Run()
```
meowgorithm, line 154:
```suggestion
	_, err := p.Run()
```

meowgorithm (changes requested):
Good catch and thank you for the PR!

clowder:
Apologies for the clumsy first attempt. I've given it a proper run through. It seems like the right thing to is to eat the `tea.ErrProgramKilled` error. Which is generated because we're closing the `tea.Program` via the context, rather than calling `Quit`.

Unfortunately I wasn't able to write a reliably failing test case for my TTY-less execution example. I'm not sure it's possible to fake a non-terminal when running tests from within a terminal.

PR #252, opened 2024-05-27
fix(filtering): resolve conflict between select and filter

MikaelFangel:
Fixes https://github.com/charmbracelet/gum/issues/580

### Changes
 - Disable the ability to toggle while filtering because they conflict and potentially cause a panic.


maaslalani:
Thank you so much @MikaelFangel!

PR #253, opened 2024-05-27
fix: remove useless output field in form struct

nervo:
Because as far as i can see, well... it's useless :)


maaslalani:
Thanks @nervo, I will merge this for now since we aren't using it but I think we'll have to reintroduce it back if we use this output for accessible mode, etc...

PR #254, opened 2024-05-27
fix(multiselect): check if is filterable before enable filtering

### LINKED ISSUE #251: MultiSelect's filterable set to "false" will have no effect.
**Describe the bug**
MultiSelect's filterable set to "false" will have no effect.

**To Reproduce**

1. run a following code:
```go
package main

import (
)

func main() {
	var types []string

	form := huh.NewMultiSelect[string]().
		Filterable(false).
		Options(
			huh.NewOption("Module", "Module").Selected(true),
			huh.NewOption("Constant", "Constant").Selected(true),
			huh.NewOption("Method", "Method").Selected(true),
			huh.NewOption("Attribute", "Attribute"),
			huh.NewOption("Class", "Class").Selected(true),
		).
		Title("Types").
		Value(&types)

	huh.NewForm(huh.NewGroup(form)).Run()

}

```
2. Click on "/"
3. Provide a "constant" value

**Expected behavior**
Filtering should not be possible on a MultiSelect field because `Filterable(false)` method was used.

 - OS: Manjaro/Linux
 - Version github.com/charmbracelet/huh v0.4.2, go 1.22

MikaelFangel:
Fixes #251 

### Changes: 
 - Check if the multi select is filterable before changing the state to filtering.
 - Update the test accordingly to the change

### DIFF
--- a/field_multiselect.go
+++ b/field_multiselect.go
@@ -522,6 +522,7 @@ func (m *MultiSelect[T]) WithTheme(theme *Theme) Field {
 // WithKeyMap sets the keymap of the multi-select field.
 func (m *MultiSelect[T]) WithKeyMap(k *KeyMap) Field {
 	m.keymap = k.MultiSelect
+	m.keymap.Filter.SetEnabled(m.filterable)
 	return m
 }
 
--- a/huh_test.go
+++ b/huh_test.go
@@ -478,7 +478,20 @@ func TestMultiSelect(t *testing.T) {
 		t.Error("Expected cursor to be on Bar.")
 	}
 
+	field.Filterable(true)
+	f = NewForm(NewGroup(field))
+	f.Update(f.Init())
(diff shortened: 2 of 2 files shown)


maaslalani:
We should actually simply disable the keybinding so it doesn't show up in the help, disabling it will also prevent this switch from activating.

MikaelFangel:
I noticed by that this will default to disable filterable should we and thus making this a breaking change. Should I add a line enabling filterable per default?

skatkov:
Awesome fix, looking forward for this to get merged ;-)

caarlos0:
> @maaslalani - I noticed by this it will default to disable filterable and thus making this a breaking change. Should we add a line enabling filterable per default?

i think so yeah... can probably do it in `New`

caarlos0:
done in https://github.com/charmbracelet/huh/pull/335/files

PR #256, opened 2024-05-28
feat: add a key getter func to options

whlit:
Add a method to get the key from the value to generate the options.
```go
type People struct {
    Name string
    Age  int
}
NewSelect[People]().Options(ToOptions(func(p People) string { return p.Name }, 
    People{Name: "Foo", Age: 20}, 
    People{Name: "Bar", Age: 25}, 
    People{Name: "Baz", Age: 23},
)...)
```

### DIFF
--- a/huh_test.go
+++ b/huh_test.go
@@ -433,6 +433,58 @@ func TestSelect(t *testing.T) {
 	}
 }
 
+func TestSelectOfInterface(t *testing.T) {
+	type People struct {
+		Name string
+		Age  int
+	}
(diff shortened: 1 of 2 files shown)


maaslalani:
Hmmmm, looking at this now. I think we probably should have made `Option` an interface!

maaslalani:
Thanks for the PR. This is a great helper function. I think to keep the API surface area of Huh? as small as possible I'd recommend we don't add this to core huh? but rather people can use this as a copy / paste helper in their own code / forms! Thank you so much for the work / contribution here. ❤️

PR #257, opened 2024-05-28
fix: apply and group styles

nervo:
I've just realized that both Form and Group styles are presents in the `Theme` struct, but none of them is applied in its respective View method.

PR #258, opened 2024-05-28
feat: form and group dedicated styles

nervo:
This pr follows this one: https://github.com/charmbracelet/huh/pull/257

Instead of having only one style for form and groups in the theme, we introduce two new structs `FormStyles` and `GroupStyles` on the same model as `FieldStyles`. Both of them having a `Base` style applied to the View.

Please note that `ErrorMessage` have also been moved from `FieldStyles` to `GroupStyles`, as its only used in group view, and in a way that suggests there was a kind of mistake :)

Note also that since both `Form` and `Group` theme struct fields are unused today, we *may* consider that it's not a breaking change.

### DIFF
--- a/form.go
+++ b/form.go
@@ -11,6 +11,7 @@ import (
 	"github.com/charmbracelet/bubbles/key"
 	"github.com/charmbracelet/bubbles/paginator"
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
 )
 
 const defaultWidth = 80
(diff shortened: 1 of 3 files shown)


maaslalani:
@nervo, I think we shouldn't break the `Theme` API unless we have a super compelling reason to. I propose we simply start using the `Form` and `Group` styles in this PR instead of changing them to `FormStyles` and `GroupStyles`

maaslalani:
If we need `Error` / `ErrorMessage` style we can add that as a separate field on the theme.

nervo:
@maaslalani using `Form` and `Group` styles is exactly (and only) the scope of this pr: https://github.com/charmbracelet/huh/pull/257 (which i just rebased)

If you're ok with it, please, consider validating and merging it :)

After that we can discuss the case of the error style on this thread.

PR #259, opened 2024-05-28
feat: add cursor text style

nervo:
'just because it's missing :)

### DIFF
--- a/theme.go
+++ b/theme.go
@@ -58,6 +58,7 @@ type FieldStyles struct {
 // TextInputStyles are the styles for text inputs.
 type TextInputStyles struct {
 	Cursor      lipgloss.Style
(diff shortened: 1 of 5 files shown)


theme.go
```diff
@@ -58,6 +58,7 @@ type FieldStyles struct {
 // TextInputStyles are the styles for text inputs.
 type TextInputStyles struct {
 	Cursor      lipgloss.Style
+	CursorText  lipgloss.Style
```
maaslalani, line 61: Do we need to add default styles or does this work with the default?

nervo: Same as `Cursor`, work as-is :)

maaslalani:
Awesome work @nervo! Thank you for catching this!

PR #260, opened 2024-05-29
feat: adjust input width to char limit

nervo:
When using non transparent background for char limited input text, the field weirdly fill all the layout view, resulting in a poor user interface experience.

In this example, `CharLimit` is set to `12`:
Capture d’écran 2024-05-29 à 14 57 29

This pr adjusts its final width during the view time, letting the `WithWidth` method doing the hard job

And voila:
Capture d’écran 2024-05-29 à 15 07 35

And before you asks, yes it supports inline mode:
Capture d’écran 2024-05-29 à 15 22 37

### DIFF
--- a/field_input.go
+++ b/field_input.go
@@ -252,6 +252,11 @@ func (i *Input) View() string {
 	i.textinput.Cursor.Style = styles.TextInput.Cursor
 	i.textinput.TextStyle = styles.TextInput.Text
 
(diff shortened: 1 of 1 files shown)


maaslalani:
Incredible work @nervo, thank you so much! I also really appreciate the attention to detail with inline mode screenshots too!

PR #261, opened 2024-05-29
feat: unselect selector

nervo:
Introduce a new style for unselected select options.

You may ask why ?

Well, before 😞 :

Capture d’écran 2024-05-29 à 19 53 55

And, after 😍  :

Capture d’écran 2024-05-29 à 19 53 23

With only
```golang
	t.Focused.SelectSelector = lipgloss.NewStyle().SetString("● ")
	t.Focused.UnselectSelector = lipgloss.NewStyle().SetString("○ ")
```

⚠️ i'm not totally satisfied by the name `UnselectSelector`... Any opinions here ?

### DIFF
--- a/field_select.go
+++ b/field_select.go
@@ -381,11 +381,18 @@ func (s *Select[T]) choicesView() string {
 		return sb.String()
 	}
 
+	// Harmonize cursors width
+	selectedCursor := styles.SelectSelector.String()
+	unselectedCursor := styles.UnselectSelector.String()
+	cursorWidth := max(lipgloss.Width(selectedCursor), lipgloss.Width(unselectedCursor))
+	selectedCursor = styles.SelectSelector.Width(cursorWidth).String()
+	unselectedCursor = styles.UnselectSelector.Width(cursorWidth).String()
+
 	for i, option := range s.filteredOptions {
 		if s.selected == i {
-			sb.WriteString(styles.SelectSelector.String() + styles.SelectedOption.Render(option.Key))
+			sb.WriteString(selectedCursor + styles.SelectedOption.Render(option.Key))
 		} else {
-			sb.WriteString(styles.UnselectSelector.String() + styles.Option.Render(option.Key))
+			sb.WriteString(unselectedCursor + styles.Option.Render(option.Key))
 		}
 		if i < len(s.options)-1 {
 			sb.WriteString("\n")


maaslalani:
I would perhaps use `Cursor` terminology here: SelectedCursor & UnselectedCursor?

maaslalani:
Not 100% if we want this as well since this makes it look like you can select multiple options.

nervo:
@maaslalani it's just a use case as an example :) None of these pseudo radio buttons are in this pull request, in facts, this is fully backward compatible, as the base "`UnselectSelector`" is "` `".

joaogmguimaraes:
This modification results in a breaking change.
Previously, `strings.Repeat` supported customizations of `SelectSelector` with sizes larger than 2 chars.

nervo:
@joaogmguimaraes i see what you mean. If you set `SelectSelector` to a size different than 2 chars, and don't touch `UnselectSelector`, then you break your view:
```
  foo
--> bar
  baz
```

In general, you will always got the same issue when `SelectSelector` and `UnselectSelector` are not the same chars size :(

We can address this by programmatically adding white spaces at the end of the shorter one. What do you think ? Is it worth it ?

joaogmguimaraes:
I think so.

When `UnselectSelector` is smaller than `SelectSelector`, we should always make them equal to maintain the existing functionality. Now, if `SelectSelector` is smaller, I don't see the need to adjust the size (although I think it is recommended to maintain consistency).

joaogmguimaraes:
In this library: [survey](https://github.com/AlecAivazis/survey?tab=readme-ov-file#changing-the-icons)

Users can add icons of different sizes, but in our case, since there is already the functionality to keep the `UnselectSelector` the same size as the `SelectSelector`, we should at least maintain this functionality working.

nervo:
@joaogmguimaraes i've just pushed a naive implementation of cursors width harmonization

@maaslalani about your renaming proposal
`SelectSelector` -> `SelectedCursor`
`UnselectSelector` -> `UnselectedCursor`
Is that ok for you ? What about current `SelectSelector`, should we deprecate it ?

maaslalani:
I think we may actually want to name this `UnselectedPrefix`? What do you think @nervo?

nervo:
@maaslalani i guess you mean the new `UnselectSelector` should be named `UnselectedPrefix`, am i right ?

In this case, what about the current `SelectSelector`, should we deprecate it in favor of something like `SelectedPrefix` ?

PR #263, opened 2024-05-30
feat: introduce accessor

nervo:
This pr introduce a new (and simple) `Accessor` interface to enhance the form field values handling in a fully backward compatible way.

This interface is generic and contains only two method `Set` (for - guess what - setting value) an `Get` (you got it, for getting value).

The `Value` method of fields is very limited, and give only one way to expose fields values - via a variable pointer - with no other possibilities.
~~I propose to deprecate it in favor of an `Accessor` to define well... the field accessor :)~~
I propose a new `Accessor` method to define well... the field accessor :) Under the hood, `Value` method still exists but makes usage of an accessor.

Two generics accessors are already available:
- `EmbeddedAccessor` which replace the current behavior of fields when no variable pointer has been defined via `Value` method.
- `PointerAccessor` wich replace the current behavior of fields when a variable pointer is defined via `Value` method.

```golang
// Good ol' fashion
.Value(&foo)
// New shiny one
.Accessor(NewPointerAccessor(&foo))
```

Custom accessors can now be defined by end users :)

### DIFF
--- a/accessor.go
+++ b/accessor.go
@@ -1,0 +1,44 @@
+package huh
+
+// Accessor give read/write access to field values.
+type Accessor[T any] interface {
+	Get() T
+	Set(value T)
+}
+
+// EmbeddedAccessor is a basic accessor, acting as the default one for fields.
+type EmbeddedAccessor[T any] struct {
+	value T
+}
+
+// Get gets the value.
+func (a *EmbeddedAccessor[T]) Get() T {
+	return a.value
+}
+
+// Set sets the value.
+func (a *EmbeddedAccessor[T]) Set(value T) {
+	a.value = value
+}
+
+// PointerAccessor allows field value to be exposed as a pointed variable.
+type PointerAccessor[T any] struct {
+	value *T
+}
+
+// NewPointerAccessor returns a new pointer accessor.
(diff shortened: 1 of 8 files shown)


maaslalani:
Hey I actually like the implementation more here but let's not deprecate `Value` as that API is much easier to use.

I like how currently the Value method simply is a shorthand for the current way of a pointer accessor.

nervo:
@maaslalani ok, let's undeprecate Value, fine 👍 
What about wording ? Do you agree with `Accessor`, `EmbeddedAccessor` and `PointerAccessor` ?

Another point that bugs me is error handling :( In a perfect world, a setter/getter should return errors, but i can't see a way to handle them without breaking the current interfaces...

maaslalani:
> @maaslalani ok, let's undeprecate Value, fine 👍 What about wording ? Do you agree with `Accessor`, `EmbeddedAccessor` and `PointerAccessor` ?
> 
> Another point that bugs me is error handling :( In a perfect world, a setter/getter should return errors, but i can't see a way to handle them without breaking the current interfaces...

You can have a `Validator` that wraps an accessor which does the error handling I believe.

maaslalani:
Fantastic work @nervo. Thank you for this!

nervo:
Wow, that was fast ❤️

Issue #166, opened 2024-03-06
Form Timeouts


korpa:
**Is your feature request related to a problem? Please describe.**
I want a `Confirm` field automatically go on with the preselection after 5 seconds without user interaction. 

**Describe the solution you'd like**
Add a timeout config to the `Run` func

Some pseudo code:

```go
form := huh.NewForm(...)
form.Run(cfg{timeout: time.Duration})
```

```go
huh.NewInput().
...
Run(cfg{timeout: time.Duration})
```

If a timeout is set, a countdown starts when Run is called. When it finishes, `Run` does not block anymore and adds preselections to Values. 
As soon as a key is hit while countdown has not finished, count down stops and works as w/o timeout

**Describe alternatives you've considered**
--

**Additional context**
--

maaslalani:
Hey! Thanks for creating this issue. We do actually need something like this as we recently removed timeout functionality for `gum` and would love to add it back due to the `huh?` transition.

Delta456:
Hi, I am working on this and I would like a complete test case/example of this.

maaslalani:
Hi @Delta456,

We would need to implement something like the following on the form, similar to the other options.

```go
func (f Form) WithTimeout(d time.Duration) Form
```

Then, when we `Run()` the form, we should pass a context to the `tea.Program` and cancel that context (use `context.WithTimeout`)

Delta456:
This needs to be closed as it is implemented now.

PR #276, opened 2024-06-08
feat(form): implement WithTimeout

Delta456:
Implements https://github.com/charmbracelet/huh/issues/166

### Demonstration

image

### DIFF
--- a/form.go
+++ b/form.go
@@ -32,6 +32,9 @@ const (
 // ErrUserAborted is the error returned when a user exits the form before submitting.
 var ErrUserAborted = errors.New("user aborted")
 
+// ErrTimeout is the error returned when the timeout is reached.
+var ErrTimeout = errors.New("timeout")
+
 // Form is a collection of groups that are displayed one at a time on a "page".
 //
 // The form can navigate between groups and is complete once all the groups are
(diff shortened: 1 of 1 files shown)


maaslalani:
Hey @Delta456, thanks for the good work: I don't have push access to your repo so can you cherry pick the following commit in this PR and then we can merge:

* https://github.com/charmbracelet/huh/commit/8ebd7d69b5986894df066ad6d1ad3a096b656dbf

This commit just adds a specific error for timeouts and handles the case when timeout is not set.

Delta456:
> Hey @Delta456, thanks for the good work: I don't have push access to your repo so can you cherry pick the following commit in this PR and then we can merge:
> 
>     * [8ebd7d6](https://github.com/charmbracelet/huh/commit/8ebd7d69b5986894df066ad6d1ad3a096b656dbf)
> 
> 
> This commit just adds a specific error for timeouts and handles the case when timeout is not set.

Done!

Forgot to consider the latter case

korpa:
@Delta456 Thank you very much. I tested it and I really like it. I have two ideas to improve it:
1) Show a countdown 
2) Would it be possible to change the behaviour a little bit? Stop timeout as soon as a key is pressed. Otherwise it could happen, that the form aborts while I'm still filling it out. Especially in bigger forms.

What do you think?

Delta456:
> @Delta456 Thank you very much. I tested it and I really like it. I have two ideas to improve it:
> 1) Show a countdown 
> 2) Would it be possible to change the behaviour a little bit? Stop timeout as soon as a key is pressed. Otherwise it could happen, that the form aborts while I'm still filling it out. Especially in bigger forms.
> 
> What do you think? 

Hi, thanks for the appreciation! I can try adding more if @maaslalani agrees too

1. It's a good idea to show a countdown. What would be the measured time stamp?

2. Exiting the timeout would beat the whole purpose of having a timeout. I feel extending the timeout would be better in bigger forms.

form.go
```diff
@@ -590,10 +602,19 @@ func (f *Form) Run() error {
...
+	if f.timeout > 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), f.timeout)
+		defer cancel()
+		f.teaOptions = append(f.teaOptions, tea.WithContext(ctx))
+	}
 	m, err := tea.NewProgram(f, f.teaOptions...).Run()
 	if m.(*Form).aborted {
 		err = ErrUserAborted
 	}
+	if err != nil && time.Since(startTime) >= f.timeout {
```
Sculas, line 615: Shouldn't this be `err == nil`? Also, every time I abort the form, I get a timeout error instead. This is because when `f.timeout` is zero, `time.Since(startTime)` will always be greater.

Delta456: It has to be `err != nil` because `tea.NewProgram` returns `ErrProgramKilled` . Though good catch!

PR #277, opened 2024-06-09
Adds a new pressable button field

andrewbenton:
The button field is styled in a similar to a confirm field with the following differences:

1. The title is the button's title rather than having a separate title and button label.
2. When pressed, the button immediately executes the configured action.  This is useful for executing actions that make immediate modifications to the form.
3. When configured to be inline, the description shows up inline after the button.

### DIFF
--- a/button.go
+++ b/button.go
@@ -60,10 +60,9 @@ func (b *Button) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 // View renders the button
 func (b *Button) View() string {
-	var (
-		styles  = b.activeStyles()
-		builder strings.Builder
-	)
+	styles := b.activeStyles()
+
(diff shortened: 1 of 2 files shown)


andrewbenton:
Here's an example of the new button on asciinema.  I couldn't find the utility that runs those `.tape` scripts.

[asciicast](https://asciinema.org/a/h08HPgzrxVnhcnHY6WYcgvwYY)

maaslalani:
Hey there! Thank you so much for the awesome PR and idea. I can see this being something we add to huh in the future. But, to keep it on the simpler side for now I'd like to not add this new field (rather perhaps we encourage this to be a field people can implement themselves or a community field, etc...).

Thank you for the great work and thank you for using Huh?

PR #285, opened 2024-06-18
refactor(input): Skip validation when going to previous field

Sculas:
In my opinion, this would make user experience better. Say you made a typo in the previous field but you already went to the next one, now you *need* to fill the current field correctly before you can fix the previous field.

This also keeps validation intact, because the field still needs to be valid to progress to the next field.


caarlos0 (approved):

caarlos0:
Thanks!

PR #287, opened 2024-06-19
fix: Fix aborting a form returning a timeout error

Sculas:
This PR fixes an issue caused by #276 where aborting a form would return `ErrTimeout` instead, because when `f.timeout == 0`, `time.Since(startTime) >= f.timeout` is **always true**. This was discussed here: https://github.com/charmbracelet/huh/pull/276#discussion_r1646225089

This is fixed by simply checking if the error returned by `bubbletea` is `ErrProgramKilled`, which is already handled by `tea.WithContext(ctx)`, making the (erroneous) check unnecessary anyway.

I've also added tests for handling timeouts, user aborts, and making sure `ErrUserAborted` is returned instead of `ErrTimeout` when a timeout error occurs (the issue mentioned above). ~~This required an extra field `program` for testing to hold the `tea.Program` to send messages to the `Form` (`batchUpdate` cannot be used here because it's managed by `tea.Program`).~~ `RunWithContext(context.Context)` has been added to support this.

### DIFF
--- a/form.go
+++ b/form.go
@@ -35,6 +35,9 @@ var ErrUserAborted = errors.New("user aborted")
 // ErrTimeout is the error returned when the timeout is reached.
 var ErrTimeout = errors.New("timeout")
 
+// ErrTimeoutUnsupported is the error returned when timeout is used while in accessible mode.
+var ErrTimeoutUnsupported = errors.New("timeout is not supported in accessible mode")
+
 // Form is a collection of groups that are displayed one at a time on a "page".
 //
 // The form can navigate between groups and is complete once all the groups are
@@ -58,9 +61,6 @@ type Form struct {
 	// purposes, if true, the form will render with basic prompting primitives
 	// to be more accessible to screen readers.
 	accessible bool
-
-	// the program this form is running in, used for testing
-	program *tea.Program
 
 	quitting bool
 	aborted  bool
@@ -93,9 +93,6 @@ func NewForm(groups ...*Group) *Form {
 		teaOptions: []tea.ProgramOption{
 			tea.WithOutput(os.Stderr),
 		},
-
-		SubmitCmd: tea.Quit,
-		CancelCmd: tea.Quit,
 	}
 
 	// NB: If dynamic forms come into play this will need to be applied when
@@ -592,6 +589,14 @@ func (f *Form) View() string {
 
 // Run runs the form.
 func (f *Form) Run() error {
+	return f.RunWithContext(context.Background())
+}
+
+// RunWithContext runs the form with the given context.
+func (f *Form) RunWithContext(ctx context.Context) error {
+	f.SubmitCmd = tea.Quit
+	f.CancelCmd = tea.Quit
+
 	if len(f.groups) == 0 {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


maaslalani:
Hey @Sculas,

I'm not able to reproduce the example where a user aborting the form causes a timeout error. Do you mind giving a minimal reproducible example?

On the burger form, when I `ctrl+c` I get a `UserAborted` error. When I have a timeout and `ctrl+c` I get a `UserAborted` error. When I have a timeout and wait, I get a `Timeout` error.

Sculas:
@maaslalani Are you sure you **don't** have a timeout set? As the PR mentions, this only occurs when `f.timeout == 0` (so no timeout set).

form.go
```diff
@@ -59,6 +59,9 @@ type Form struct {
 	// to be more accessible to screen readers.
 	accessible bool
 
+	// the program this form is running in, used for testing
+	program *tea.Program
```
maaslalani, line 63: Can we keep the program out of the form. We can send an `Update` to the form which allows the same keypress to be sent during testing.

form.go
```diff
@@ -90,6 +93,9 @@ func NewForm(groups ...*Group) *Form {
 		teaOptions: []tea.ProgramOption{
 			tea.WithOutput(os.Stderr),
 		},
+
+		SubmitCmd: tea.Quit,
+		CancelCmd: tea.Quit,
```
maaslalani, line 98: Let's keep these in `Run` since this was intentional to avoid having these pre-set in embedded forms in bubble tea applications.

huh_test.go
```diff
@@ -843,6 +846,31 @@ func TestSkip(t *testing.T) {
...
+		t.Errorf("expected timeout error, got %v", err)
+	}
+
+	// Test that the form aborts without throwing a timeout error when explicitly told to abort.
+	// run() is blocking, so we need to send the message in a separate goroutine.
+	go func() {
+		// Give it some time to start.
+		time.Sleep(200 * time.Millisecond)
+		// Tell the form to abort.
+		f.program.Send(tea.KeyMsg{Type: tea.KeyCtrlC})
```
maaslalani, line 865: We can use `f.Update(tea.KeyMsg{Type: tea.KeyCtrlC})`

maaslalani:
Sorry you're totally right, I was on an incorrect branch!

form.go
Sculas: The handler for `keymap.Quit` then returns a `tea.Quit` cmd to exit the program, but that won't happen if I use something like `batchUpdate`. What would the correct way be to handle this?

maaslalani: A form is actually simply a bubble tea model, so we can create a new tea program and just run that here and pass the form model to the program.

You would also need to set the quitCmd etc...

Sculas: I see, that works. However, the intention was to test the `Run()` method, because that's where the issue was. Creating a new program inside the test would bypass the `Run()` method, no?

Sculas:
I've applied the changes you requested and removed the `program` field by introducing `RunWithContext(context.Context)`, which made sense anyway because the timeout used `context.Background()` instead of accepting an existing context.
Of course, `Run()` is kept and just calls `RunWithContext(context.Background())` as it did previously as well.

I've fixed the data race as well and all tests now pass with `-race`.

Sculas:
@maaslalani Would you be able to look into this soon?

maaslalani:
@Sculas This looks great! Thank you!

Issue #196, opened 2024-04-25
Spinner: Context ignored/dropped under Accessible mode


polds:
**Describe the bug**
The `Context()` method only passes the context to the underlying bubbletea program. If you set the Spinner to use Accessible mode then the context isn't used.

Also worth noting, even though `Run()` returns an error, it is not possible to have a non-nil error: https://github.com/charmbracelet/huh/blob/acfe24c3f5b5cc857596778d7885051f9a0d19c1/spinner/spinner.go#L138-L169 

So I do see the expected behavior of fast cancellation on context cancellation, just with no error. But when in Accessible mode the context is ignored.

**To Reproduce**

https://go.dev/play/p/MTobz4rLf7t
```golang
package main

import (
	"context"
	"testing"
	"time"

	"github.com/charmbracelet/huh/spinner"
)


func TestContextCancellation(t *testing.T) {
        ctx := context.Background()
	t.Run("works", func(t *testing.T) {
		ctx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
		t.Cleanup(cancel)

		var i int
		// Note: Spinner does not correctly bubble up the deadline exceeded error, but it
		//  does correctly stop execution.
		_ = spinner.New().Action(func() {
			time.Sleep(100 * time.Millisecond)
			i++
		}).Context(ctx).Run()
		if i != 0 {
			t.Errorf("Got count of %d, but expected 0", i)
		}
	})
	t.Run("bug", func(t *testing.T) {
		ctx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
		t.Cleanup(cancel)

		var i int
		// Note: Spinner does not correctly bubble up the deadline exceeded error, but it
		//  does correctly stop execution.
		_ = spinner.New().Action(func() {
			time.Sleep(100 * time.Millisecond)
			i++
		}).Context(ctx).Accessible(true).Run()
		if i != 0 {
			t.Errorf("Got count of %d, but expected 0", i)
		}
	})
}
```

**Expected behavior**
- `Run()` returns a context.Cancelled error type
- `Accessible(true)` still handles context cancellation

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

maaslalani:
Hey @polds, thanks so much for the issue!

This is now fixed (let me know if there's any problems though):

https://github.com/charmbracelet/huh/blob/main/spinner/spinner.go#L160-L192

testinfected:
[...]

PR #292, opened 2024-06-24
feat(spinner): improve context, action, output, tests

caarlos0:
This PR started as a small one but grew up as I was finding more problems.

- setting both an action and a context caused the context to be ignored
- context is now always set if there's any action
- added `ActionWithError(func(context.Context) error)`, which allows to use the spinner context and return an error, improving error cases usage
- added `Output` to set the output, mainly for testing, but might be useful in more places
- added more tests, many of them testing both accessible and regular flows
- when running in the regular flow, the action is now run within bubbletea loop (as a `tea.Cmd`) instead of outside
- added more spinner examples 
- updated go module to 1.22


spinner/spinner.go
```diff
@@ -165,6 +166,9 @@ func (s *Spinner) Run() error {
 	if errors.Is(err, tea.ErrProgramKilled) {
 		return nil
 	} else {
+		if actionErr != nil {
```
ccoVeille, line 169: No need for an else above as you return.

Code could be simpler and more readable

spinner/examples/loading/main.go
```diff
@@ -8,9 +8,13 @@ import (
 )
 
 func main() {
-	action := func() {
-		time.Sleep(2 * time.Second)
+	action := func() error {
```
ccoVeille, line 11: I usually recommend passing a context on such function


```suggestion
	action := func(ctx context.Context) error {

       select {
         case <- ctx.Done:
             return nil
         case <- time.After(2 * time.Second):
            return nil
       }
```

This way the code can react to a canceled event that occurs before the end if the time.Sleep

spinner/spinner.go
```diff
@@ -98,7 +98,7 @@ func New() *Spinner {
 	s.Style = lipgloss.NewStyle().Foreground(lipgloss.Color("#F780E2"))
 
 	return &Spinner{
-		action:     func() { time.Sleep(time.Second) },
+		action:     func() error { time.Sleep(time.Second); return nil },
```
ccoVeille, line 101: You should create a NewTimer

```go
func NewTimer(ctx context, d time.Duration) func(context.Context) error {
   return func(ctx context.Context) error {
            select { 
                 case <- ctx.Done: 
                 return nil 
                 case <- time.After(2 * time.Second): 
                 return nil
            }
}
``

This way you can reuse it. Here I'm using a context as described before, but you can also do without

File: spinner/spinner.go [hunk 4]
```diff
@@ -153,10 +153,11 @@ func (s *Spinner) Run() error {
 		return s.ctx.Err()
 	}
 
+	var actionErr error
 	p := tea.NewProgram(s, tea.WithContext(s.ctx), tea.WithOutput(os.Stderr))
 	if s.ctx == nil {
 		go func() {
-			s.action()
+			actionErr = s.action()
 			p.Quit()
```
ccoVeille, line 161: I don't understand how the code there could send an error to a code that is out of the go func, so without having a channel and wait for it.

There is something strange to me in the existing code. And I'm afraid the changes are making things worse

ccoVeille: BTW, I hope `p.Quit` doesn't call `os.Exit`otherwise the code out of the go routine (I'm already doubtful about) will never be executed

caarlos0:
another idea would be to have a 

```go
type Spinner[T any] struct {
  action func () T
  // ..
```

and maybe a

```go
type Spinner2[T any, Q any] struct {
  action func() (T, Q)
}
```


the later being most commonly used as `Spinner2[Something, error]` I would guess, maybe even worth making the second arg an error always.

spinner/examples/loading/main.go
caarlos0: there are other examples using contexts

caarlos0:
@maaslalani I reworked some stuff:

- run the `action` as part as a `tea.Cmd`
- improve handling when having only ctx, only action, or both

spinner/spinner.go
caarlos0: i rewrote this in the latest commit

spinner/spinner.go
caarlos0: this no longer exists :)

spinner/examples/loading/main.go
ccoVeille: My point was about passing it to the action

As you seem confident it's a good thing to use a time.Sleep, but maybe it's only for example purpose

ccoVeille (approved):

caarlos0:
I think this might also fix #196

caarlos0:
one bad thing about this, is that it might be a breaking change 🤔

ccoVeille:
Can't you support both signature with generics? 🤔

spinner/spinner.go
```diff
@@ -23,12 +22,14 @@ import (
 // ⣾  Loading...
 type Spinner struct {
 	spinner    spinner.Model
-	action     func()
+	action     func() error
```
ccoVeille, line 25: I would expect to receive either the context or spinner as argument.


```suggestion
	action     func(ctx context.Context) error
```


```suggestion
	action     func(s *Spinner) error
```

It would help. I only expect to get the context. But maybe the spinner could be useful

spinner/spinner.go
```diff
@@ -61,7 +62,7 @@ func (s *Spinner) Title(title string) *Spinner {
 }
 
 // Action sets the action of the spinner.
-func (s *Spinner) Action(action func()) *Spinner {
+func (s *Spinner) Action(action func() error) *Spinner {
```
ccoVeille, line 65: The only issue you have with breaking changes is here.


```suggestion
func (s *Spinner) Action(action func()) *Spinner {
      return s.ActionErr(func() {
              action()
              return nil
    })
}

func (s *Spinner) ActionErr(action func()) *Spinner {
```

ccoVeille (approved):
LGTM 👍

caarlos0:
also added the feature from #172 here

spinner/examples/context-and-action-and-error/main.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	err := spinner.New().
+		Context(ctx).
+		ActionErr(func(context.Context, io.Writer) error {
+			time.Sleep(time.Minute)
```
ccoVeille, line 20: Even it's an example, I won't use a time. Minute here

You can use a 5s delay it would be enough.

I know this timeout won't be reached but three is a problem with a time.Sleep that will block everything. It's somehow OK for an example. But as it's an example, it should be written in a production-ready code

So here is what I'm suggesting

```suggestion
			select {
				case <-time.After(5 * time.Second):
				case <-ctx.Done():
			}
```

ccoVeille (changes requested):

spinner/examples/printing/main.go
```diff
@@ -0,0 +1,30 @@
...
+		fmt.Fprintln(w, "Added top bun")
+		time.Sleep(time.Second)
+		return nil
+	}
+	_ = spinner.New().
+		Title("Preparing your burger").
+		ActionErr(action).
+		// Accessible(true).
+		Run()
+	fmt.Println("Order up!")
```
ccoVeille, line 29:
```suggestion
	var w bytes.Buffer

	action := func(_ context.Context) error {
		fmt.Fprintln(w, "Added bottom bun")
		time.Sleep(time.Second)
		fmt.Fprintln(w, "Added patty")
		time.Sleep(time.Second)
		fmt.Fprintln(w, "Added condiments")
		time.Sleep(time.Second)
		fmt.Fprintln(w, "Added top bun")
		time.Sleep(time.Second)
		return nil
	}
	_ = spinner.New().
		Title("Preparing your burger").
		ActionErr(action).
		// Accessible(true).
		Run()
	fmt.Println("Order up!")
```

spinner/spinner.go
```diff
@@ -23,12 +24,15 @@ import (
 // ⣾  Loading...
 type Spinner struct {
 	spinner    spinner.Model
-	action     func()
+	action     func(ctx context.Context, w io.Writer) error
```
ccoVeille, line 27: I'm unsure about this

Some may need a spinner, most won't. Context is there to be able to cancel things

Maybe, if you want to do return string, don't add a buffer in the signature and in the spinner, but in the returned value

So something like

```suggestion
	action     func(ctx context.Context) (string, error)
```

Or refactor everything to pass the spinner itself as argument

ccoVeille (changes requested):

spinner/spinner.go
caarlos0: its not to return the string, its to print above the spinner

spinner/spinner.go
```diff
@@ -62,6 +66,15 @@ func (s *Spinner) Title(title string) *Spinner {
...
 // Action sets the action of the spinner.
 func (s *Spinner) Action(action func()) *Spinner {
+	s.action = func(context.Context, io.Writer) error {
+		action()
+		return nil
+	}
+	return s
+}
+
+// ActionErr sets the action of the spinner.
```
bashbunni, line 76: We should specify what the difference is between `Action` and `ActionErr`

bashbunni:
I think I'm missing something obvious here - what's wrong with how we're currently handling actions and how does this change improve that?

caarlos0:
> I think I'm missing something obvious here - what's wrong with how we're currently handling actions and how does this change improve that?

if i recall correctly, if you passed both a context and a function it didn't work properly, that's fixed here

spinner/spinner.go
caarlos0: done

meowgorithm (approved):
Wow, nice work!

spinner/spinner.go
```diff
@@ -62,6 +62,18 @@ func (s *Spinner) Title(title string) *Spinner {
...
+//
+// This is just like [Action], but allows the action to use a `context.Context`
+// and to return an error.
+func (s *Spinner) ActionErr(action func(context.Context) error) *Spinner {
```
meowgorithm, line 76:
```suggestion
func (s *Spinner) ActionWithErr(action func(context.Context) error) *Spinner {
```

spinner/examples/context-and-action-and-error/main.go
caarlos0: other examples also don't have it like this... the focus of the example is on the spinner, the sleeps are just to mimic something that would take some time... 

gonna leave the simple thing, I did changed the timeout to 5s though :)

Thanks for the review!

spinner/spinner.go
```diff
@@ -60,8 +61,27 @@ func (s *Spinner) Title(title string) *Spinner {
...
+	s.action = func(context.Context) error {
+		action()
+		return nil
+	}
+	return s
+}
+
+// ActionWithErr sets the action of the spinner.
+//
+// This is just like [Action], but allows the action to use a `context.Context`
```
ccoVeille, line 82:
```suggestion
// This is just like [Spinner.Action], but allows the action to use a `context.Context`
```

It's a method on a struct, not a function

spinner/spinner.go
```diff
@@ -60,8 +61,27 @@ func (s *Spinner) Title(title string) *Spinner {
 	return s
 }
 
+// Output set the output for the spinner.
+// Default is STDOUT when [Accessible], STDERR otherwise.
```
ccoVeille, line 65:
```suggestion
// Default is STDOUT when [Spinner.Accessible], STDERR otherwise.
```

spinner/spinner_test.go
```diff
@@ -105,10 +116,63 @@ func TestSpinnerUpdate(t *testing.T) {
...
+			tb.Errorf("expected error to be %v, got %v", target, err)
+		}
+	}
+}
+
+func requireContextCanceled(tb testing.TB, err error) {
+	tb.Helper()
+	if !errors.Is(err, context.Canceled) && !errors.Is(err, tea.ErrProgramKilled) {
+		tb.Errorf("expected to get a context canceled error, got %v", err)
 	}
```
ccoVeille, line 177: A switch could be more readable


```suggestion
	switch {
		case errors.Is(err, context.Canceled):
		case errors.Is(err, tea.ErrProgramKilled):

		default:
			tb.Errorf("expected to get a context canceled error, got %v", err)
 	}
```

ccoVeille: I'm surprised you said 👍 but didn't apply while you did with my other 2 feedbacks.

But this one is not yet resolved, so I might simply wait 😅

ccoVeille (approved):

PR #304, opened 2024-07-11
Refactored option rendering in field_select.go to improve distinction

csandeep:
Fixed rendering of unselected options in the select field. Previously, all options were displayed with the same style regardless of their selection state. Now, unselected options will be appropriately rendered with an 'UnselectedOption' style. 

ref: https://github.com/charmbracelet/gum/issues/614


meowgorithm (approved):
LGTM!

PR #313, opened 2024-07-16
Fix 'Dynamic forms' Page link in README

MaximilianSoerenPollak:
Hi, 
I just noticed that the 'dynamic forms' page link in the README isn´t working. 

I just made a quick fix so that it now works again :) 

Let me know if anything else is needed.


maaslalani:
Thank you so much ❤️

PR #325, opened 2024-07-24
fix(select): select all/none

caarlos0:
refs https://github.com/charmbracelet/gum/issues/550


field_multiselect.go
```diff
@@ -377,6 +382,26 @@ func (m *MultiSelect[T]) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				}
 			}
 			m.updateValue()
+		case key.Matches(msg, m.keymap.ToggleAll) && !m.filtering && m.limit == 0:
```
aymanbagabas, line 385: Can we use this while filtering?

caarlos0: oh yeah, my bad

PR #331, opened 2024-07-25
fix: set YOffset + cursor based on first selected item

bashbunni:
image
tested against gum @ `main-huh` (basically just made it depend on this branch of huh)
image

examples all work normally

you can test with Gum (after a lil go get -u github.com/charmbracelet/huh@refactor-select) then try seq 22 | go run .  choose --selected=2,21 --limit=2 (or set limit to one or none for single select)

### DIFF
--- a/field_multiselect.go
+++ b/field_multiselect.go
@@ -127,17 +127,24 @@ func (m *MultiSelect[T]) Options(options ...Option[T]) *MultiSelect[T] {
 		return m
 	}
 
+	minSelected := len(options) - 1
 	for i, o := range options {
-		for _, v := range m.accessor.Get() {
-			if o.Value == v {
-				options[i].selected = true
-				break
+		if o.selected {
+			options[i].selected = true
+			if i < minSelected {
+				minSelected = i
 			}
+			break
 		}
+	}
+	// No selected items... Start cursor at the beginning.
+	if minSelected == len(options)-1 {
+		minSelected = 0
 	}
 	m.options.val = options
 	m.filteredOptions = options
 	m.updateViewportHeight()
+	m.cursor = minSelected
 	return m
 }
 
@@ -549,11 +556,20 @@ func (m *MultiSelect[T]) optionsView() string {
 	return sb.String()
 }
 
+// centerYOffsetToSelected sets the YOffset so the selected element is in the
+// middle of the list.
+func (m *MultiSelect[T]) centerYOffsetToSelected() {
+	if m.cursor > m.viewport.YOffset+m.viewport.Height {
+		// Start with the selected value in the middle of the viewport.
+		m.viewport.SetYOffset(m.cursor - (m.viewport.Height+m.viewport.YOffset)/2)
+	}
+}
+
 // View renders the multi-select field.
 func (m *MultiSelect[T]) View() string {
 	styles := m.activeStyles()
-
 	m.viewport.SetContent(m.optionsView())
+	m.centerYOffsetToSelected()
 
 	var sb strings.Builder
 	if m.title.val != "" || m.title.fn != nil {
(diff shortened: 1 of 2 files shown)


maaslalani (approved):

maaslalani:
Nice! This is a great UX improvement!

bashbunni:
Fixes: https://github.com/charmbracelet/gum/issues/606

meowgorithm:
Totally, would almost call this a feature versus a fix. @bashbunni, can you add example Go code we can use to vet this?

bashbunni:
@meowgorithm Just added some instructions. I mainly tested it with Gum. 
1. go to gum and replace the huh dep with this branch
2. `seq 22 | go run . choose --selected=2,21 --limit=2`

caarlos0 (approved):

meowgorithm:
We should test with `huh` the library itself as well. It looks like there's still a bug when navigating through groups with <kbd>tab</kbd>. Example here:

```go
package main

import (
	"fmt"
	"os"

)

func main() {
	f := huh.NewForm(
		huh.NewGroup(
			huh.NewSelect[string]().
				Options(
					huh.NewOption("A", "a"),
					huh.NewOption("B", "b"),
					huh.NewOption("C", "c"),
					huh.NewOption("D", "d"),
					huh.NewOption("E", "e"),
					huh.NewOption("F", "f"),
					huh.NewOption("G", "g"),
					huh.NewOption("H", "h"),
					huh.NewOption("I", "i"),
					huh.NewOption("J", "j"),
					huh.NewOption("K", "k").Selected(true),
					huh.NewOption("L", "l"),
					huh.NewOption("M", "m"),
					huh.NewOption("N", "n"),
					huh.NewOption("O", "o"),
					huh.NewOption("P", "p"),
				),
		).WithHeight(8),
		huh.NewGroup(
			huh.NewMultiSelect[string]().
				Options(
					huh.NewOption("A", "a"),
					huh.NewOption("B", "b"),
					huh.NewOption("C", "c"),
					huh.NewOption("D", "d"),
					huh.NewOption("E", "e"),
					huh.NewOption("F", "f"),
					huh.NewOption("G", "g"),
					huh.NewOption("H", "h"),
					huh.NewOption("I", "i"),
					huh.NewOption("K", "k").Selected(true),
					huh.NewOption("L", "l"),
					huh.NewOption("M", "m"),
					huh.NewOption("N", "n"),
					huh.NewOption("O", "o").Selected(true),
					huh.NewOption("P", "p"),
				),
		).WithHeight(8),
	)

	if err := f.Run(); err != nil {
		fmt.Fprintf(os.Stderr, "Oof: %v", err)
	}
}
```

bashbunni:
Going to test more edge cases before merging this one :)

meowgorithm:
Yeah, please do look at the example I posted above as well

maaslalani:
You may also need to test with Dynamic Huh? That is, what happens when options are swapped out and some are selected.

bashbunni:
The group-related issue is actually caused by a different bug I found and documented [here](https://github.com/charmbracelet/huh/issues/349). `WithHeight` wasn't setting the viewport height and was instead just cutting content. That's why the calculations in this branch weren't working properly. Will continue trying to break it though :)

bashbunni:
Maybe I'm doing something wrong, but I can't seem to get pre-selected options to work in dynamic huh. You can try for yourself by checking out the `dynamic-selection-example` branch and running the `dynamic/dynamic-selections` example.
https://github.com/charmbracelet/huh/blob/dynamic-selection-example/examples/dynamic/dynamic-selections/main.go

I created this example to test pre-selected elements in dynamic huh, but the cursor always starts at the top of the options anyway. Let me know if this is by design!

Issue #399, opened 2024-08-31
Allow Cursor option for the filepicker


radar07:
**Is your feature request related to a problem? Please describe.**
It's nice to allow the user to set the cursor for the file picker. (Default is `>`).

**Describe the solution you'd like**
When the user sets the custom cursor option like the one below,

![image](https://github.com/user-attachments/assets/37991066-3315-4bc7-8c36-bb3e76830929)

the output could be something like the below
![image](https://github.com/user-attachments/assets/0050c160-23f9-4e18-8ea4-e57a6543d2fd)

**Additional context**
I'm unsure if this is a feature or a bug because the gum file-picker allows the option to set the cursor but it isn't working. Please see this [issue](https://github.com/charmbracelet/gum/issues/662)

![image](https://github.com/user-attachments/assets/0222264f-0af4-4d0b-b53d-993488382348)

radar07:
I'd love to raise a PR. Thanks:)

PR #400, opened 2024-09-01
feat: add cursor option to filepicker

radar07:
This might close https://github.com/charmbracelet/huh/issues/399.

Further, we could use this in the [gum](https://github.com/charmbracelet/gum) file picker.


caarlos0:
thanks!

PR #407, opened 2024-09-10
feat: Add Cursor() methods to Select and MultiSelect fields

kralicky:
Hi! This PR addresses a small limitation I ran into while building out a UI with this library. I was using a MultiSelect field to display a list of options, and wanted to use a Note to show additional details/context information for the option under the cursor. However, there was no way to know which option that was.

With this change, you can now write the following:

```go
items := []string{...}

ms := huh.NewMultiSelect[string]().
	Options(huh.NewOptions(items...)...)

huh.NewNote().
	DescriptionFunc(func() string {
		currentItem := items[ms.Cursor()]
		// do something with currentItem
	}, onCursorUpdate{ms}) // <- see below
```

There is still the small issue of knowing when to invalidate the DescriptionFunc - I solved this by using the value of `Cursor()` as a hash:

```go
type onCursorUpdate struct {
	Field interface{ Cursor() int }
}

// Hash implements hashstructure.Hashable
func (u onCursorUpdate) Hash() (uint64, error) {
	return uint64(u.Field.Cursor()), nil
}
```

This relies on implementation details however, so if this use case is common, something like it (but with a more idiomatic api) could be a good candidate for a built-in utility.

### DIFF
--- a/field_multiselect.go
+++ b/field_multiselect.go
@@ -220,6 +220,17 @@ func (m *MultiSelect[T]) Blur() tea.Cmd {
 	return nil
 }
 
+// Hovered returns the value of the option under the cursor, and a bool
+// indicating whether one was found. If there are no visible options, returns
+// a zero-valued T and false.
+func (m *MultiSelect[T]) Hovered() (T, bool) {
+	if len(m.filteredOptions) == 0 || m.cursor >= len(m.filteredOptions) {
+		var zero T
+		return zero, false
+	}
+	return m.filteredOptions[m.cursor].Value, true
+}
+
 // KeyBinds returns the help message for the multi-select field.
 func (m *MultiSelect[T]) KeyBinds() []key.Binding {
 	binds := []key.Binding{
--- a/field_select.go
+++ b/field_select.go
@@ -281,6 +281,17 @@ func (s *Select[T]) Blur() tea.Cmd {
 	return nil
 }
 
+// Hovered returns the value of the option under the cursor, and a bool
+// indicating whether one was found. If there are no visible options, returns
+// a zero-valued T and false.
+func (m *Select[T]) Hovered() (T, bool) {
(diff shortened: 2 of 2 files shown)


Siri-chan:
+1 because I actually also have this use-case, more or less - but with a `map[string]string` rather than a `[]string`.

Some comments on the change:
Wouldn't this make more sense to return the `[T comparable]` key that the MultiSelect is built with?
Just exporting `cursor` as readonly doesn't seem to fit the way the rest of the API returns items from the list of options (see `MultiSelect.Value(*[]T)`, for example - it doesn't just return the indices of the options, but rather their keys).
Maybe something of this form would be more flexible and appropriate?
```go
func (m *MultiSelect[T]) Hovered() T {
	return m.options.val[m.cursor]
}
```

kralicky:
@Siri-chan I agree, good idea. Plus, the lists can be filtered, so just having the cursor index might not be sufficient to know what option is hovered. 

Here's what I came up with that works for my use case, what do you think:
```go
// Hovered returns the value of the option under the cursor, and a bool
// indicating whether one was found. If there are no visible options, returns
// a zero-valued T and false.
func (m *MultiSelect[T]) Hovered() (T, bool) {
	if len(m.filteredOptions) == 0 || m.cursor >= len(m.filteredOptions) {
		var zero T
		return zero, false
	}
	return m.filteredOptions[m.cursor].Value, true
}
```

Siri-chan:
Seems fine by me - your branch works for my use-case and doesn't break any of the demo tapes as best I can.

kralicky:
After some more testing, it looks like there is a possible race condition here. Commands are evaluated in a separate goroutine from the one that handles messages, and the DescriptionFunc/TitleFunc/etc functions are called during command evaluation in the background, so they can't read anything that could be modified in another component's Update(). Not sure the best way to fix this.

caarlos0:
Thanks!

Issue #296, opened 2024-06-26
Hiding implicit focus indicators


meowgorithm:
Having seen Huh in action for a bit, I think a good improvement would be to eliminate the focus indicators when they're implicit, with the goal of making the experience more visually focused and concise.

In other words, I don't think we always need to draw the little bar that indicates the active section in all cases.

Off the top of my head, the logic would work like this:

1. If the entire form is one single field, hide the indicator.
2. If the entire form has multiple groups with one single field in each group, hide all indicators.
3. In all other cases show the indicator.

dez11de:
Creators of such forms that don't want the little bar can simply set theme.Focused.Base.BorderStyle to HiddenBorder.

meowgorithm:
This is somewhat true, but from a design perspective I'd prefer to default to not showing implicit indicators to reduce visual noise.

There's also a catch: focus indicators may not necessarily be borders: a theme could instead use a background color on the group, for example, so manually adjusting borders won't necessarily do the trick.

bashbunni:
Hey @meowgorithm I've converted this to a discussion so we can talk through it and keep our backlog clear. If it does end up being something we want to add, we'll promote it back to an issue when we are ready to work on it.

I'm not sure when the focus indicator would be implicit, maybe it would be good to give an example case for future reference.

PR #423, opened 2024-09-30
feat!: v2

caarlos0:
- [x] update bubbletea v2
- [x] update bubbles v2
- [ ] update wish
- [ ] other breaking changes we might need to do


bashbunni:
I don't know how this impacts #296 but I'm converting that feature to a discussion as we aren't ready to go ahead with it yet. Maybe something we address down the line when Huh is already on v2?

Issue #417, opened 2024-09-17
huh.NewConfirm() proceeds even with validation error


istiak101:
**Describe the bug**
huh.NewConfirm() proceeds even with validation error while huh.NewSelect [string] () doesn't proceed to next option unless errors are nil.

**To Reproduce**
```
huh.NewConfirm().
	Title("Install ABC?").
	Value(&two).
	Validate(func(install bool) error {
	    if !install {
		return errors.New("You have to install this!")
	    }
	    return nil
	}),

huh.NewConfirm().
	Title("Install XYZ?").
	Value(&three),
```

**Expected behavior**
huh.NewConfirm() shouldn't proceed like huh.NewSelect [string] () if validation returns error.

prithvijj:
Just adding relevant examples and minimal code to maybe explain the issue?

Add this in `example/test1/main.go` which is to showcase the issue above
```
package main

import (
	"errors"

	"github.com/charmbracelet/huh"
)

func main() {
	// var commit, scope string
	// var summary, description string
	var confirm bool

	huh.NewForm(
		huh.NewGroup(
			huh.NewConfirm().
				Title("Install ABC?").
				Value(&confirm).
				Validate(func(install bool) error {
					if !install {
						return errors.New("you have to install this")
					}
					return nil
				}),
			huh.NewConfirm().
				Title("Install XYZ?").
				Value(&confirm).
				Validate(func(install bool) error {
					if !install {
						return errors.New("you have to install this")
					}
					return nil
				}),
		),
	).Run()

}

```
Then running
```
cd examples/test1
go run *.go 
```

Normal view
![image](https://github.com/user-attachments/assets/7b3452d8-de20-4250-96c2-78864fb1e462)

Pressing `enter` on the negative option results in the validate function being triggered, however it moves to the next field in the group, which is unintended since the Validate function would have not allowed to move to the next field
![image](https://github.com/user-attachments/assets/e46335db-7e42-4605-ac39-bdee9d4a8255)

Pressing `enter` on the negative option for the second confirm field, results in the error message again that's on the side, which seems to be weird?
![image](https://github.com/user-attachments/assets/82591f5c-e629-4164-a4ef-d3e581535d94)


Also just doing `go run *.go` and Pressing the Left Key moves both the confirm fields, which seems weird too, since we expect that the first confirm field is being worked on, hence it should be in focus/interactable

![image](https://github.com/user-attachments/assets/95724788-bad8-4925-a696-a473dadbe0f2)

---

[...]

PR #428, opened 2024-10-17
fix(confirm): Ensure validation occurs when moving within fields

prithvijj:
## Changes

- Added in a fix to ensure that when the confirm field has a Validate method provided, it should validate the value, and does not move between different fields
- Previously when a user submits the confirm field, it would just move to the next field in the `huh.Group`, reason being that within the `Update()` method of `field_confirm`, the `NextField()` tea.Cmd would be used, but there was no checks/Validation taking place to see if it's allowed to move to the next field
- Hence now, if the user submits the confirm field, it ensures the Validate method is called, to see if the value is allowed, and then allow to move to the next field

## Testing Notes

```
package main

import (
	"errors"
	"fmt"

)

func main() {
	var confirm bool
	var confirm2 bool

	huh.NewForm(
		huh.NewGroup(
			huh.NewConfirm().
				Title("Install ABC?").
				Value(&confirm).
				Validate(func(install bool) error {
					if !install {
						return errors.New("you have to install this")
					}
					return nil
				}),
			huh.NewConfirm().
				Title("Install XYZ?").
				Value(&confirm2).
				Validate(func(install bool) error {
					if !install {
						return errors.New("you have to install this")
					}
					return nil
				}),
		),
	).Run()

	fmt.Printf("confirm: %v confirm2: %v\n", confirm, confirm2)
}
```

Previously:

Rendering two confirm fields, which have validate
image

Pressing `Enter` ends up moving the selection from the first field, to the second one (even though the validate function is called, but doesn't block from moving to the new field)

image

Pressing `Enter` again ends up moving the selection the next field, but doesn't block from moving to the next field
image

Now:

Rendering two confirm fields which have validate
image

Pressing `Enter` does not move the selection to the next field, and blocks it (due to the validate function)
image

Pressing `Left`
image

Pressing `Enter` ends up moving the selection to the next field, since the validation passed
image

Pressing `Enter` again does not move the selection to the next field and blocks it (due to the validate function)
image

Works as exepcted
image

## Other Notes

Related to: https://github.com/charmbracelet/huh/issues/417
- [ ] Anything else missing, What permutations could I test out?

### DIFF
--- a/field_confirm.go
+++ b/field_confirm.go
@@ -204,14 +204,30 @@ func (c *Confirm) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 			c.accessor.Set(!c.accessor.Get())
 		case key.Matches(msg, c.keymap.Prev):
+			c.err = c.validate(c.accessor.Get())
+			if c.err != nil {
+				return c, nil
+			}
 			cmds = append(cmds, PrevField)
 		case key.Matches(msg, c.keymap.Next, c.keymap.Submit):
+			c.err = c.validate(c.accessor.Get())
+			if c.err != nil {
+				return c, nil
+			}
 			cmds = append(cmds, NextField)
 		case key.Matches(msg, c.keymap.Accept):
 			c.accessor.Set(true)
+			c.err = c.validate(c.accessor.Get())
+			if c.err != nil {
+				return c, nil
+			}
 			cmds = append(cmds, NextField)
 		case key.Matches(msg, c.keymap.Reject):
 			c.accessor.Set(false)
+			c.err = c.validate(c.accessor.Get())
+			if c.err != nil {
+				return c, nil
+			}
 			cmds = append(cmds, NextField)
 		}
 	}


bashbunni (approved):

prithvijj:
> This looks great and this behaviour makes sense to me. That said, I'm not the most familiar with this code base, so I'll get another team member to sign off on this too before merging :)

Yepp, sounds goood!

bashbunni:
Hey @prithvijj thanks so much for your patience with this. After discussing with the team, this is actually the desired behaviour. It turns out we've modelled Gum forms to be quite similar in design to HTML forms which allow you to continue past an invalid field, but you must fix it before submitting the form.

Thank you so much for your contribution here, but we will keep the current form behaviour as-is

prithvijj:
> Hey @prithvijj thanks so much for your patience with this. After discussing with the team, this is actually the desired behaviour. It turns out we've modelled Gum forms to be quite similar in design to HTML forms which allow you to continue past an invalid field, but you must fix it before submitting the form.
> 
> Thank you so much for your contribution here, but we will keep the current form behaviour as-is

No worries @bashbunni ! Thanks for taking a look at it :tada: 
Was about to link your comment in the original issue, but can't seem to find that issue which I swear I remember linking on my PR description, oh well :shrug:

Issue #446, opened 2024-10-31
Writing a todo list form


muhmud:
**Is your feature request related to a problem? Please describe.**
I'm trying to write a simple todo list form, with a select above, and an input below. I would like to have the input focused when the form group loads, but haven't been able to achieve this, after much effort.

I would also like to customize the behaviour that occurs when I press enter on the input, so that it adds the item to the todo list and keeps the form alive. I can achieve this by reloading the form, and this is mostly OK, but it would be nice to have more control over this.

**Describe the solution you'd like**
* You should be able to choose which field of a group has focus when a group is loaded
* You should be able to control whether submit actually submits or whether the submit is cancelled

**Describe alternatives you've considered**
I've tried to call `Update()` on a group directly with `NextField()` to try to get it to move to the next field, which works, but the select shows the message `Loading...`

**Additional context**
Seems like this kind of thing should be simple to achieve

bashbunni:
Hey @muhmud this seems to be a question related to usage, so I'm going to go ahead and convert this to a discussion.

> You should be able to control whether submit actually submits or whether the submit is cancelled

You can change what happens on submit by creating your own [`tea.Cmd`](https://github.com/charmbracelet/bubbletea/tree/main/tutorials/commands) for the form's [`SubmitCmd`](https://pkg.go.dev/github.com/charmbracelet/huh#Form).

Also if the user presses `ctrl+c` it will run the `CancelCmd` for the form.

> You should be able to choose which field of a group has focus when a group is loaded

Typically focus happens in order of declaration. Could you please help me understand a case where that might not be desired?

---

As you mentioned, a reload would be needed to populate new values onto a list field. If you want more control over that, then migrating your project to use Bubble Tea instead will give you more control. I hope this helps! Please let me know if you have further questions or concerns.

PR #469, opened 2024-11-17
feat: add option to not quit form after submission

jonas-grgt:
Adds the option to keep the form visible after submission. Afterwards the form could be *reactivated* using `f.Init()`.
This is useful for when submitting something that might fail and needs some further editing.


jonas-grgt:
This should close #446

jonas-grgt:
Maybe it would be better to expose this functionality through an explicit setter instead of directly exposing the field?

As in:

```go
form.WithQuitAfterSubmit(false); // to not quit the form after submitting
```

WDYT?

jonas-grgt:
Just rebased this PR to resolve conflicts and keep it up to date.
Would love to get some feedback or a quick review when you have a moment, happy to make any changes needed to get this merged in. Thanks for your time!

PR #473, opened 2024-11-19
feat: allow Input to be prefilled with an initial string literal

jonas-grgt:


caarlos0:
Hey! Thanks for the PR!

For now, I think the current behavior is good enough. To actually use the input you'll have to set the value anyway, so it makes sense to me to the default value in said value variable instead.

That said, closing this.

Cheers!

jonas-grgt:
You do not need to set the value and can just rely on the getter methods of the form. In that case having to first initialize a variable to be able to pass a pointer to the Value method is clunky.

PR #477, opened 2024-11-20
feat: expose input field Messages

jonas-grgt:
Enables standalone use of input fields by exposing the messages emitted by its commands


bashbunni:
Hey @jonas-grgt what kind of workflow would this change enable? Working with the fields in your own Bubble Tea application? It would be great if you could provide a usage example on this one 🙏

jonas-grgt:
@bashbunni  correct, it would enable direct field usage without the need for a Form.

PR #479, opened 2024-11-21
fix: wrapped content in multi-select

bashbunni:
- **feat: set group with from size msg**
- **feat: handle WindowSizeMsg + base height on rendered content**
- **feat: render prefix before newlines**
- **feat: update viewport on resize, cursor != height so no scroll...**
- **fix: always update viewport height**


bashbunni:
Need to fix select too. Also lost the bullet point and checkmark before the elements.

Before:
image

After:
image

bashbunni:
image

group.go
```diff
@@ -290,6 +291,16 @@ func (g *Group) fullHeight() int {
 	return height
 }
 
+// width returns the full width of the group.
```
jonas-grgt, line 294: nitpick: change // width to // fullWidth

jonas-grgt (approved):

Issue #429, opened 2024-10-18
Select options disappear when other options wrap


seeseemelk:
**Describe the bug**
When showing a Select menu, the last option disappears when at least two options have to wrap

**To Reproduce**
1. Create a select with at least three (long) options
1. Make your terminal small enough so that two options have to wrap
2. See the last one disappear

**Expected behavior**
All possible options should stay visible.

**Screenshots**
No wrapping, all options are visible:
![Screenshot_2024-10-18_07-51-19](https://github.com/user-attachments/assets/6c9bcddc-583f-4dd6-8f3d-c13401a63b38)

One option wraps, all options are visible:
![Screenshot_2024-10-18_07-51-29](https://github.com/user-attachments/assets/077baef5-1288-4a46-a756-336694164146)

Two options wrap, the last option disappears:
![Screenshot_2024-10-18_07-51-42](https://github.com/user-attachments/assets/09f4b87f-f079-4d5f-a16c-1fec96776e54)

**Desktop (please complete the following information):**
 - OS: `Linux dalek 6.11.3-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 10 Oct 2024 20:11:06 +0000 x86_64 GNU/Linux`
 - Terminal: alacritty, cool retro term, urxvt

seeseemelk:
I've been looking into this a little: the select option hard-codes that it should scroll down a single line when pressing the down arrow key.
Changing this to, say, 2, helps a little, but does not actually fix it.

The Viewport class itself also seems to assume that lines don't wrap. It generates a slice of visible lines based on the viewport height and scroll position, but this slice does not take wrapping into account.

Likely, if the Viewport is modified so it correctly considers wrapping, the Select will probably *just work*

bashbunni:
I agree this is likely `viewport` not calculating height properly

Kyren223:
@bashbunni It is, I independently also discovered, there is an open issue at bubbles https://github.com/charmbracelet/bubbles/issues/644

caarlos0:
this will be fixed with bubbles v2 and viewport horizontal scroll

Issue #398, opened 2024-08-31
v0.5.3 breaks Select.Title when options don't fit on the screen


adsr303:
**Describe the bug**

**To Reproduce**
1. Clone https://github.com/adsr303/manup. Note: Depends on `whatis`.
2. `go run .` - the title "Choose manpage" should be visible.
3. `go get github.com/charmbracelet/huh@latest`
4. `go mod tidy`
5. `go run .` - the title will not be visible.
6. Add `manOpts = manOpts[:5]` in main.go:30.
7. `go run .` - now all options fit in the screen and the title "Choose manpage" is visible.

This is a regression from v0.5.2.

**Expected behavior**
The title of the `Select` should be visible whether the options fit on the screen or not.

**Screenshots**

**Desktop (please complete the following information):**
 - OS:  Ubuntu 22.04.4 LTS on Windows 10 x86_64
 - Kernel: 5.15.153.1-microsoft-standard-WSL2
 - Terminal: Windows Terminal, Tilix

**Additional context**
This may be somehow related to another observation: When any of the items don't fit on one line and wrap around, the title also doesn't appear.

seeseemelk:
This seems related to my issue: https://github.com/charmbracelet/huh/issues/429

Though I can still get the options to break once a lot (read: all) options have to wrap

caarlos0:
@seeseemelk probably yes 

closing as a duplicate

PR #485, opened 2024-11-26
fix: wrapping and height calculations

bashbunni:
## Changes

Content used to cut off if the height of the window was smaller than the content height. 
Wrapped content broke formatting and caused content to get cut off. 
The offset calculations were off and wouldn't scroll to show hidden content.
There was an off by one error for content that exceeded the window height (would always hide the title of the first field).

## Other considerations
Some fieds (confirm, select, input) have the option to be shown as inline. When shown as inline, the styles do not wrap (intentional)

Buttons will stack vertically if they don't fit horizontally. Button widths will also match (do we want that by default or only when stacked?)

## Fixes
closes https://github.com/charmbracelet/huh/issues/398
closes https://github.com/charmbracelet/huh/issues/429

Maybe others (need to confirm)

Here's what this PR looks like:
https://github.com/user-attachments/assets/b3588782-33d4-4e11-8105-2f6b7335df6a

Previous behaviour:
https://github.com/user-attachments/assets/0e77808e-5895-45d4-b075-f299babecf2b

A few enhancements I'd still like to see: huh does a lot of redraws that cause rendering issues when not in alt screen. You come across that behaviour when resizing the window a lot while running an example (e.g. burger)
---
- **feat: set group with from size msg**
- **feat: handle WindowSizeMsg + base height on rendered content**
- **feat: render prefix before newlines**
- **feat: update viewport on resize, cursor != height so no scroll...**
- **fix: always update viewport height**
- **fix: add option prefix**
- **fix(multiselect): width calculation**
- **fix: calculate offset from previous field heights**
- **feat: add calculate wrapping helper**
- **fix(select): wrap select field**
- **feat: make group set field widths accounting for styles**
- **feat: set width**
- **feat: stack buttons in narrow window**
- **feat: add Theme function to Fields**
- **feat: calculate width based on field theme**
- **fix(group): fix height calculation**
- **feat: remove FullWidth helper (not needed)**
- **fix(group): fix off by one at top for overflowed content**


form.go
```diff
@@ -521,7 +524,7 @@ func (f *Form) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 		f.selector.Range(func(_ int, group *Group) bool {
 			if group.fullHeight() > msg.Height {
-				group.WithHeight(msg.Height)
+				group.WithHeight(msg.Height - 1)
```
bashbunni, line 527: this accounts for the help menu at the bottom of the window taking up one row. Can add a const for this 1 for legibility if you'd prefer

group.go
```diff
@@ -285,25 +288,26 @@ func (g *Group) fullHeight() int {
 	height := g.selector.Total()
 	g.selector.Range(func(_ int, field Field) bool {
 		height += lipgloss.Height(field.View())
+		height += lipgloss.Height(gap)
```
bashbunni, line 291: without this line, content that fits in the current window size will truncate its height slightly

what it looks like without this line:
https://github.com/user-attachments/assets/7a3a0d33-2524-49ca-b334-02ba072af10d

bashbunni:
In case we don't want to merge the stacked buttons here, I have another version of this branch where I removed those features. https://github.com/charmbracelet/huh/tree/horizontal-btns

bashbunni:
rip, looks like these changes might be breaking grid layout...

image

In the image above we're missing options 4-6. This is happening at all widths. Absolutely tragic 😭 

Build info:
macOS, kitty, no multiplexer, just running the layouts/grid example

field_select.go
```diff
@@ -514,16 +514,22 @@ func (s *Select[T]) updateViewportHeight() {
...
 }
 
+// Theme returns the theme of the field.
+func (s *Select[T]) Theme() *Theme {
+	theme := s.theme
+	if theme == nil {
+		theme = ThemeCharm()
+	}
+	return theme
+}
```
caarlos0, line 531:
```suggestion
	if s.theme != nil {
		return s.theme
	}
	return ThemeCharm()
}
```

feels a bit better I think 🤔

group.go
```diff
@@ -268,7 +270,8 @@ func (g *Group) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	switch msg := msg.(type) {
 	case tea.WindowSizeMsg:
-		g.WithHeight(max(g.height, min(g.fullHeight(), msg.Height-1)))
+		g.WithHeight(min(g.fullHeight(), msg.Height-1))
```
caarlos0, line 273: what is this -1 here?

bashbunni:
The grid rendering issue may be related to https://github.com/charmbracelet/bubbletea/issues/573...

The fourth to sixth elements are shown when I print `form.View`, but for some reason when running the form it's cutting it off in the terminal window. I only see this in the terminal, but not when logging the output at all. You can see this by running `go test -run TestLayoutGrid -v` locally. That's the same code as the `layout/grid` example.

I thought it might be related to `layoutGrid.visibleGroups`, but I ruled it out (removed the condition + made it all visible)

group.go
bashbunni: I think that's the help menu, just added a const to make this more clear

group.go
```diff
@@ -268,7 +273,8 @@ func (g *Group) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	switch msg := msg.(type) {
 	case tea.WindowSizeMsg:
-		g.WithHeight(max(g.height, min(g.fullHeight(), msg.Height-1)))
+		g.WithHeight(min(g.fullHeight(), msg.Height-helpMenuHeight))
+		g.WithWidth(msg.Width)
```
bashbunni, line 277: This is what's breaking grid layout. Changes seem to work fine without it, so can remove it

testdata/TestLayoutGrid.golden
```diff
@@ -0,0 +1,23 @@
+  First   Fourth     
+  >       >
```
bashbunni, line 2: delete this file

caarlos0:
let me know if you can take a look at it and if you find any new bugs hehe

Issue #512, opened 2025-01-19
Option to disable EDITOR for huh.Text


ArchiMoebius:
If huh is used in tandem with wish, using huh.Text exposes the server to RCE (for example, EDITOR=vim, then the user simply has to type `:! rm -rf /`

**Describe the solution you'd like**
Request boolean to disable editor.

**Describe alternatives you've considered**
A docker-compose sandbox would mitigate the issue - to a degree, better to have an option to completely turn off EDITOR however.

caarlos0:
fixed in #516 

thanks!

PR #516, opened 2025-01-28
feat: option to disable external editor in Text field

rahji:
Adds the ability to disable the external editor option in text fields. I was looking for this feature when I noticed that someone else had asked for the same thing in issue #512

### DIFF
--- a/huh_test.go
+++ b/huh_test.go
@@ -374,6 +374,31 @@ func TestText(t *testing.T) {
 	}
 }
 
+func TestTextExternalEditorHidden(t *testing.T) {
+	field := NewText().ExternalEditor(false)
+	f := NewForm(NewGroup(field))
+	f.Update(f.Init())
+
+	// Type Huh in the form.
+	m, _ := f.Update(keys('H', 'u', 'h'))
+	f = m.(*Form)
+	view := ansi.Strip(f.View())
+
+	if !strings.Contains(view, "Huh") {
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
hey! thanks for the PR!

would you mind adding a test for it as well?

rahji:
Yes, I can do that 👍🏽

rahji:
I added a test and squashed the commits.

huh_test.go
```diff
@@ -374,6 +374,31 @@ func TestText(t *testing.T) {
...
+
+	if !strings.Contains(view, "Huh") {
+		t.Log(pretty.Render(view))
+		t.Error("Expected field to contain Huh.")
+	}
+
+	if !strings.Contains(view, "alt+enter / ctrl+j new line • enter submit") {
+		t.Log(pretty.Render(view))
+		t.Error("Expected field to contain help, without ctrl+e.")
+	}
```
caarlos0, line 395: probably safer to check that view doesn't not contain 'ctrl+e'?

jonas-grgt (approved):

huh_test.go
rahji: I was trying to mirror the other text field test, but I agree with you. I updated it.

caarlos0:
thanks!

PR #518, opened 2025-01-29
chore: use bubbletea v2 generics

aymanbagabas:
This makes Huh use the new generics API defined here https://github.com/charmbracelet/bubbletea/pull/1298


caarlos0:
afaik we can close this, let me know otherwise ayman 🙏🏻

Issue #385, opened 2024-08-23
Don't render title row when title is empty


meowgorithm:
I've run into this a few times where the context for the UI is presented elsewhere and we really just need the interactive portion of the component.

jonas-grgt:
@meowgorithm  I believe this has already been implemented at least for `field_input.go` as the `View` method contains the following:

```go
	if i.title.val != "" || i.title.fn != nil {
		sb.WriteString(styles.Title.Render(i.title.val))
		if !i.inline {
			sb.WriteString("\n")
		}
	}
```

bashbunni:
Thanks @jonas-grgt, you're right this is fixed on `main`. Here's what it looks like when I make the title empty for the `Select`

With titles set:
![Image](https://github.com/user-attachments/assets/a4620329-31a9-48df-b280-a52dc1540fbd)

Here's when I set all of the titles to "". Looks like it's working with `Select`, but not `Confirm`.
![Image](https://github.com/user-attachments/assets/b14a0af7-681a-4c76-8820-623d4ebc584b)

Just added some tests for this in a PR. https://github.com/charmbracelet/huh/pull/560

PR #560, opened 2025-03-07
test: check rendering of empty titles and descriptions

bashbunni:
https://github.com/charmbracelet/huh/issues/385

Confirm is the only field not rendering properly when title is empty, as shown in the new tests. Feel free to add fix to this branch :)

### DIFF
--- a/huh_test.go
+++ b/huh_test.go
@@ -936,6 +936,150 @@ func TestGetFocusedField(t *testing.T) {
 	}
 }
 
+func TestTitleRowRender(t *testing.T) {
+	t.Run("Confirm", func(t *testing.T) {
+		withTitle := NewConfirm().Title("A Title")
+		noTitle := NewConfirm().Title("")
+		got := lipgloss.Height(ansi.Strip(noTitle.View()))
+		want := lipgloss.Height(ansi.Strip(withTitle.View())) - 1
+		if got != want {
+			t.Fatalf("got != want; height should be %d, got %d", want, got)
+		}
+	})
+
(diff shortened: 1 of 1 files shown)


huh_test.go
```diff
@@ -936,6 +936,150 @@ func TestGetFocusedField(t *testing.T) {
...
+
+func TestDescriptionRowRender(t *testing.T) {
+	t.Run("Confirm", func(t *testing.T) {
+		withDescription := NewConfirm().Description("A Description")
+		noDescription := NewConfirm().Description("")
+		got := lipgloss.Height(ansi.Strip(noDescription.View()))
+		want := lipgloss.Height(ansi.Strip(withDescription.View())) - 1
+		if got != want {
+			t.Fatalf("got != want; height should be %d, got %d", want, got)
+		}
```
caarlos0, line 1019: would probably be better to do something like:

```suggestion
		c := NewConfirm().Description("A Description")
		if view := c.View(); !strings.Contains(view, "A Description") {
			t.Log(pretty.Render(view))
			t.Error("Expected description to be visible")
		}
```

on all these tests

on a second thought, this can be a bit less repetitive as well, let me do it in #548 as I'm already fixing things related to titles/descriptions there 🙏

aymanbagabas (approved):

huh_test.go
bashbunni: @caarlos0 the only thing is that when there are empty titles, the row still renders, then when the field is selected, it accentuates that

image

In essence what I'm trying to test here is that the field height shrinks if the description and titles are not set

PR #574, opened 2025-03-15
Fixed README GetString to GetInt for bubbletea form example

hegner123:
In Readme, the example of using huh with bubbletea incorrectly uses the GetString method on the level field when this field is an Int.


raphamorim (approved):
Thank you @hegner123

PR #580, opened 2025-03-17
CORRECTED comment to match function

hegner123:
PrevField method was labeled as NextField in comment. No functional change.


andreynering (approved):
Thanks!

PR #593, opened 2025-03-19
fix(filepicker): height, navigation

### LINKED ISSUE #470: NewFilePicker().Height seems to be ignored and filepicker overtakes containing form
**Describe the bug**
1. When setting `NewFilePicker().Picking(false).Height(*someInt*)` or `NewFilePicker().Picking(false)` (no height set) or `NewFilePicker().Height(*someInt*)`, the file picker expands (when activated) to the height of its container group or form. It also opens in a "new screen" rather than existing in place in its containing form. See screenshot below.
2. When setting `NewFilePicker().Picking(true).Height(*someInt*) or `NewFilePicker().Picking(true)` (no height set), the returned height is always the full terminal height. It also opens in a "new screen" rather than existing in place in its containing form. See screenshot below.

**To Reproduce**
1. When `Picking` is false:
```go
huh.NewForm(
	huh.NewGroup(
		huh.NewNote().
			Description("life add electronics"),
		huh.NewFilePicker().
			Title("Datasheet file  ").
			Height(5).
			CurrentDirectory("../collections/electronics/files").
			ShowPermissions(false).
			DirAllowed(false).
			FileAllowed(true).
			Value(&tmpDatasheetPath),
		huh.NewInput().
			Value(&tmpFilename).
			Title("Filename"),
		huh.NewText().
			Value(&tmpNote).
			Placeholder("Best day ever.").
			Title("Optional notes").
			CharLimit(0).
			Lines(3),
	),
).Run()
```

2. When `Picking` is true:
(Code is same as above but with the addition of `Picking(true).` after `huh.NewFilePicker().`.)

**Screenshots**
1. When `Picking` is false:
   Screenshot 2024-11-18 at 8 20 36 AM Screenshot 2024-11-18 at 8 20 48 AM
   When activated, the filepicker is displayed in a new screen that is the same height as its containing form or group. 
2. When `Picking` is true:
   Screenshot 2024-11-18 at 8 18 50 AM
   When the form is displayed, the filepicker automatically overtakes the form and displays in a new screen that is the entire height of the terminal.

**Expected behavior**
1. Firstly, I didn't expect the filepicker to open in its own "window"... I thought it would operate in place like a `Select`. This would be my preference.
2. Secondly, I would expect the `Height()` to impact the rendering height of the filepicker (again, in place on the form if possible).

 - OS: Mac OS

**Additional context**
I feel like I've seen the filepicker work as expected in this version of `huh`, so I wonder if there's some additional aspect that's causing this behavior such its containing group or form.

caarlos0:
closes #470

### DIFF
--- a/keymap.go
+++ b/keymap.go
@@ -71,19 +71,19 @@ type MultiSelectKeyMap struct {
 
 // FilePickerKey is the keybindings for filepicker fields.
 type FilePickerKeyMap struct {
-	Open     key.Binding
-	Close    key.Binding
-	GoToTop  key.Binding
-	GoToLast key.Binding
-	PageUp   key.Binding
-	PageDown key.Binding
-	Back     key.Binding
-	Select   key.Binding
-	Up       key.Binding
-	Down     key.Binding
-	Prev     key.Binding
-	Next     key.Binding
-	Submit   key.Binding
+	Open       key.Binding
+	Close      key.Binding
+	GotoTop    key.Binding
+	GotoBottom key.Binding
+	PageUp     key.Binding
+	PageDown   key.Binding
+	Back       key.Binding
+	Select     key.Binding
+	Up         key.Binding
+	Down       key.Binding
+	Prev       key.Binding
+	Next       key.Binding
+	Submit     key.Binding
 }
 
 // NoteKeyMap is the keybindings for note fields.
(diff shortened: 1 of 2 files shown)


keymap.go
```diff
@@ -71,19 +71,19 @@ type MultiSelectKeyMap struct {
...
-	Select   key.Binding
-	Up       key.Binding
-	Down     key.Binding
-	Prev     key.Binding
-	Next     key.Binding
-	Submit   key.Binding
+	Open       key.Binding
+	Close      key.Binding
+	GotoTop    key.Binding
+	GotoBottom key.Binding
```
caarlos0, line 77: all other fields use this nomenclature

aymanbagabas (approved):

PR #614, opened 2025-04-03
fix: accessible mode IO, default values, and more

### LINKED ISSUE #612: Accessible prompts ignore configured stdio streams
### Describe the bug

Consider the following form:

```go
return huh.NewForm(groups...).
	WithTheme(huh.ThemeBase16()).
	WithAccessible(p.accessible).
	WithProgramOptions(tea.WithOutput(p.stdout), tea.WithInput(p.stdin))
```

We would expect that `p.stdout` and `p.stdin` are respected in accessible mode (at least, I can't think of any reason it shouldn't?). As far as I can tell however, the various `runAccessible` methods ignore this and use the currently exec-ed commands streams:

https://github.com/charmbracelet/huh/blob/310cd4a379ac64219f9188779c7d3f00378a6c03/accessibility/accessibility.go#L71-L81

https://github.com/charmbracelet/huh/blob/310cd4a379ac64219f9188779c7d3f00378a6c03/field_select.go#L711-L719

I'm not exactly sure right now whether that has an implication for `gh` production use, but it sure makes it a pain to test.

This is where the options get used, which is out of the `runAccessible` branch:

https://github.com/charmbracelet/huh/blob/310cd4a379ac64219f9188779c7d3f00378a6c03/form.go#L674

caarlos0:
- fixed I/O in accessibility mode
- text didn't echo the user input after submit (whereas the other components did)
- fixed default value on confirm in accessible mode
- fixed a non-set `CharLimit` of `0` being considered in the Text component

closes #612 
closes https://github.com/charmbracelet/huh/issues/615


accessibility/accessibility.go
```diff
@@ -16,7 +16,7 @@ import (
 // Given invalid input (non-integers, integers outside of the range), the user
 // will continue to be reprompted until a valid input is given, ensuring that
 // the return value is always valid.
-func PromptInt(prompt string, low, high int) int {
+func PromptInt(w io.Writer, r io.Reader, prompt string, low, high int) int {
```
andreynering, line 19: @caarlos0 You're changing the signature of some functions. Isn't this a problem regarding backwards compatibility?

caarlos0: but yeah, will update to keep compat

form.go
```diff
@@ -318,13 +325,17 @@ func (f *Form) WithHeight(height int) *Form {
 }
 
 // WithOutput sets the io.Writer to output the form.
+// Default is STDOUT when [Form.Accessible], STDERR otherwise.
```
bashbunni, line 328: this field isn't public, so this link doesn't work. 
```suggestion
// Default is STDOUT when [Form] is accessible (set with [Form.WithAccessible], STDERR otherwise.
```

bashbunni (approved):
There are a couple of lint complaints on the tests, but looks good to me :)

bashbunni:
I made the burger example accessible and the second text field didn't print my input, but the first one did :thinking: 
image

field_confirm.go
```diff
@@ -298,20 +300,20 @@ func (c *Confirm) View() string {
...
 }
 
 // runAccessible runs the confirm field in accessible mode.
-func (c *Confirm) runAccessible() error {
+func (c *Confirm) runAccessible(w io.Writer, r io.Reader) error {
 	styles := c.activeStyles()
 	if c.title.val != "" {
-		fmt.Println(styles.Title.Render(c.title.val))
-		fmt.Println()
+		_, _ = fmt.Fprintln(w, styles.Title.Render(c.title.val))
```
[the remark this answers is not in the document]
caarlos0: yes, one of the values is an error, this just makes clear we're ignoring it

andreynering (approved):

PR #626, opened 2025-04-14
feat(input): dynamic default values

zevdg:
Default values are a special type of placeholder that is treated as the field's actual value when the value would otherwise be empty. The existing default value mechanism — the field's "initial value" — is regularly overridden by the value read from the textinput bubble, which makes it effectively static.

It might be possible to make the "initial value" mechanism more dynamic, leveraging the placeholder for default values instead as this PR does allows dynamic defaults to be more consistent with the rest of the dynamic forms API and it also yields a slightly better user experience by allowing default values to be treated and themed differently than user-entered values.

Fixes the issue raised in https://github.com/charmbracelet/huh/discussions/582#discussioncomment-12708978.

The 2nd commit updates the themes to display default values differently than placeholders. This isn't strictly necessary, but since they behave slightly differently, they probably should look different too.

### DIFF
--- a/examples/dynamic/dynamic-default/main.go
+++ b/examples/dynamic/dynamic-default/main.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/charmbracelet/huh"
+)
+
+func main() {
+
+	var name, id, directory string
+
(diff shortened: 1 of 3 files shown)


zevdg:
Just checking in here.  Is there anything I can do to move this PR forward?

Issue #628, opened 2025-04-15
Select field: Selected option not visible on big lists


edsonmedina:
### Describe the bug

If I create a select field with a long list of options (bigger than the viewport) it becomes scrollable.
But if I pre-select an option that is not in the first viewport page I can't see the cursor.

### To Reproduce

- create a select field with a long list (ie: 50 options)
- pre-select an option down the list (ie: position 30)
- run

### Source Code

```golang 
package main

import (
        "github.com/charmbracelet/huh"
)

func main() {

        country := "PT"

        x := huh.NewSelect[string]().
                Title("Pick a country.").
                Options(
                        huh.NewOption("United States", "US"),
                        huh.NewOption("Germany", "DE"),
                        huh.NewOption("Brazil", "BR"),
                        huh.NewOption("Canada", "CA"),
                        huh.NewOption("Spain", "ES"),
                        huh.NewOption("Portugal", "PT"), // outside of initial viewport
                        huh.NewOption("Italy", "IT"),
                        huh.NewOption("Germany", "DE"),
                ).
                Height(5). // small viewport
                Value(&country) // pre-select an option below the viewport bottom
        x.Run()
}
         
```

### Expected behavior

I expect the viewport to show the cursor on the selected option

PR #629, opened 2025-04-15
Center viewport on selected option

edsonmedina:
This fixes bug #628

### DIFF
--- a/field_select.go
+++ b/field_select.go
@@ -199,7 +199,19 @@ func (s *Select[T]) selectOption() {
 			break
 		}
 	}
-	s.viewport.YOffset = s.selected
+	// Center the selected option in the viewport if possible.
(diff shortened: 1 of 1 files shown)


caarlos0:
I don't think we need all these changes 🤔 

gonna push a pr with a one line change that I think does the trick as well

caarlos0:
see #632 

ps: it wont "center" the option, just make sure it's visible

let me know what you think

edsonmedina:
> ps: it wont "center" the option, just make sure it's visible
> let me know what you think

Yeah, I tried making it visible (first option on the viewport), but IMO it made it feel like it was the first option on the list (it wasn't clear that there were more options above), but that's just personal taste, not a bug.

Thanks!

PR #642, opened 2025-04-29
fix(select,multiselect): properly handle defaults and prompts

### LINKED ISSUE #638: Accessible `Select` prompt treats no default value (`0`) as if the default is `1`
### Describe the bug

👋

The accessible `Select` prompt treats no default value (`0`) as if the default is `1`. I think it should instead treat it as invalid input by printing a message and continue the prompting loop. 

https://github.com/charmbracelet/huh/blob/6116e38ee23fd28386271434c5df5992fe8ba334/field_select.go#L721-L734

### To Reproduce

Consider the following form:

```Go
	formOptions := []huh.Option[int]{
		huh.NewOption("Option 1", 1),
		huh.NewOption("Option 2", 2),
		huh.NewOption("Option 3", 3),
	}
	prompt = "accessible Select test"
	// zero value default
	var result int
	_ = huh.NewForm(
		huh.NewGroup(
			huh.NewSelect[int]().
				Title(prompt).
				Value(&result).
				Options(formOptions...),
		),
	).
		WithTheme(huh.ThemeBase16()).
		WithAccessible(true).
		Run()
	fmt.Println("Selected option:", result)
```

Results in the following behavior (press `<enter>`, no input):

```
accessible Select test 
1. Option 1
2. Option 2
3. Option 3
Input a number between 1 and 3: 

Selected option: 1
```

### LINKED ISSUE #639: Awkward select message in accessible mode when there's only one option
### Describe the bug

> [!NOTE]
> This is not a bug report, but an improvement request we already discussed internally.

In accessibility mode, when there is a select input with only one item, the prompt message reads *"Input a number between 1 and 1"* as in the screenshot below.

This is a bit awkward to the user. However, I'm not sure this is intentional. If it's not, I think a different wording might help the user, like: *"There is only one option; enter number 1"* or *"Only one option available; enter number 1"*.

### Setup

Please complete the following information along with version numbers, if applicable.
 - Ubuntu
 - zsh
 - gnome-terminal

### To Reproduce

1. Create a Go file with this content:
   ```go
   // test-prompt.go
   package main

   import "github.com/charmbracelet/huh"

   func main() {
   	huh.NewForm(huh.NewGroup(huh.NewSelect[int]().Options(huh.Option[int]{Key: "foo"}))).WithAccessible(true).Run()
   }
   ``` 
1. Run `go run test-prompt.go`
1. See the input prompt message: *"Input a number between 1 and 1"*

### Expected behavior

A better message like *"There is only one option; enter number 1"* or *"Only one option available; enter number 1"*, or anything.

### Screenshots

Image

caarlos0:
closes #639
closes #638

### DIFF
--- a/huh_test.go
+++ b/huh_test.go
@@ -1333,7 +1333,7 @@ func TestAccessibleFields(t *testing.T) {
 			CheckOutput: func(tb testing.TB, output string) {
 				tb.Helper()
 				requireContains(tb, output, "Select: ")
-				requireContains(tb, output, "Only one option available")
+				requireContains(tb, output, "There is only one option available; enter the number 1:")
 				requireContains(tb, output, "Invalid: must be 1")
 			},
 			CheckValue: func(tb testing.TB, value any) {
--- a/huh_test.go
+++ b/huh_test.go
@@ -1298,16 +1298,47 @@ func TestAccessibleFields(t *testing.T) {
 		"select default value": {
 			FieldFn: func() Field {
 				v := "c"
-				return NewSelect[string]().Options(NewOptions("a", "b", "c", "d")...).Value(&v)
+				return NewSelect[string]().
+					Options(NewOptions("a", "b", "c", "d")...).
+					Value(&v)
 			},
 			Input: "\n",
 			CheckOutput: func(tb testing.TB, output string) {
 				tb.Helper()
 				requireContains(tb, output, "Select: ")
+				requireContains(tb, output, "Input a number between 1 and 4")
 			},
 			CheckValue: func(tb testing.TB, value any) {
 				tb.Helper()
 				requireEqual(tb, "c", value.(string))
+			},
+		},
+		"select no input": {
+			Field: NewSelect[string]().Options(NewOptions("a", "b")...),
+			Input: "\n2\n",
+			CheckOutput: func(tb testing.TB, output string) {
+				tb.Helper()
+				requireContains(tb, output, "Select: ")
+				requireContains(tb, output, "Input a number between 1 and 2")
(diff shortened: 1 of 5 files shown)


huh_test.go
```diff
@@ -1298,18 +1298,49 @@ func TestAccessibleFields(t *testing.T) {
...
-				return NewSelect[string]().Options(NewOptions("a", "b", "c", "d")...).Value(&v)
+				return NewSelect[string]().
+					Options(NewOptions("a", "b", "c", "d")...).
+					Value(&v)
 			},
 			Input: "\n",
 			CheckOutput: func(tb testing.TB, output string) {
 				tb.Helper()
 				requireContains(tb, output, "Select: ")
+				requireContains(tb, output, "Input a number between 1 and 4")
```
meowgorithm, line 1309:
```suggestion
				requireContains(tb, output, "Enter a number between 1 and 4")
```

huh_test.go
```diff
@@ -1298,18 +1298,49 @@ func TestAccessibleFields(t *testing.T) {
...
 				requireEqual(tb, "c", value.(string))
 			},
 		},
+		"select no input": {
+			Field: NewSelect[string]().Options(NewOptions("a", "b")...),
+			Input: "\n2\n",
+			CheckOutput: func(tb testing.TB, output string) {
+				tb.Helper()
+				requireContains(tb, output, "Select: ")
+				requireContains(tb, output, "Input a number between 1 and 2")
```
meowgorithm, line 1322:
```suggestion
				requireContains(tb, output, "Enter a number between 1 and 2")
```

PR #667, opened 2025-07-09
feat: make Field.RunAccessible public, deprecate Field.WithAccessible

### LINKED ISSUE #663: Change Request: Make huh.Field interface implementable again
Hi,
I already wrote on Discord about that topic. 
2 Month ago a bug fix was merged which introduced a new method to the `huh.Field` interface called [runAccessible](https://github.com/charmbracelet/huh/commit/c5906631cbb549ce70bbcd45b5d295f821e78842#diff-5b6a5ae74565d7ef9e9f413f5e0afd815515e98a7a11186b88ebe150ea90e7cdR154). 
With this addition the interface lost the possibility to be implemented by other users.
Since the charmbracelet team is very careful to add new field types it would be very helpful to be able to add your own types in your application.
Also I think making `runAccessible` public is not that much of an issue so I would like to propose to make the field public.

Cheers,
lyssar

caarlos0:
This way we can remove make the public interface Field implementable by 3rd parties again.

closes #663

### DIFF
--- a/field_confirm.go
+++ b/field_confirm.go
@@ -35,7 +35,7 @@ type Confirm struct {
 	width           int
 	height          int
 	inline          bool
-	accessible      bool
+	accessible      bool // Deprecated: use RunAccessible instead.
 	theme           *Theme
 	keymap          ConfirmKeyMap
 	buttonAlignment lipgloss.Position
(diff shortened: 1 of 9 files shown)


caarlos0:
> @caarlos0 ask me to look at his changes. They look good to me but I'm not sure about the process (should I do the review and approve it? is this done by someone else?)

if you feel like it, go for it :)

PR #725, opened 2026-01-09
fix(input): do not run validation on blur to allow backward navigation (Shift+Tab)

nelvko:
## Summary
This PR removes unconditional validation from Input.Blur() so that validation only occurs when the user explicitly attempts to advance or submit a field (e.g. Enter).

This fixes an issue where navigating backward with Shift+Tab triggers validation and prevents users from returning to previous fields, which is inconsistent with expected TUI form behavior.

## Solution
Remove validation from Input.Blur() and rely on existing validation logic already present in the Update handler for Next and Submit key bindings.

## Checklist before requesting a review
- [ ] I have created a discussion that was approved by a maintainer (for new features).

### DIFF
--- a/field_input.go
+++ b/field_input.go
@@ -254,7 +254,6 @@ func (i *Input) Blur() tea.Cmd {
 	i.focused = false
 	i.accessor.Set(i.textinput.Value())
 	i.textinput.Blur()
-	i.err = i.validate(i.accessor.Get())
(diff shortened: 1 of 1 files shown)


kienmarkdo:
Are there any updates on this fix? @meowgorithm @kujtimiihoxha

PR #728, opened 2026-01-25
fix: select cursor now moves instead of options scrolling

### LINKED ISSUE #724: NewSelect OptionsFunc cursor is static while options move
### Description

When creating a `NewSelect` with normal `Options`, I am able to display all my items while the cursor on the left moves up and down between them. The options are static. This is great.

When I create the same form with `OptionsFunc` the cursor becomes static and the options scroll up and down. This pushes the top options out of the view port.

```go
// Here is the working Options way
huh.NewSelect[string](). 
    Title("Option Working List").
    Options(
        huh.NewOption("One", "one"),
        huh.NewOption("Two", "two"),
        huh.NewOption("Three", "three"),
  )
```

Image

As you can see in the image, the second option ("Two") is selected and "One" is still visible.

```go
huh.NewSelect[string]().
	Title("OptionFunc Not working List").
	OptionsFunc(func() []huh.Option[string] {
		return huh.NewOptions("One", "Two", "Three") // For simplicity
	}, "option").
```

Image

Image to show all three are displaying.

Image

Now we select the second option, and the first option ("One") is out of the viewport.

Also note that for all of this I have set and unset the `.Height(10)` to test out if the viewport wasn't tall enough. This had no affect on the rendering issues. 

### Version

v0.8.0

### Environment

Mac Arm (M1 pro) OS Sequoia 15.6.1 / Ghostty & Wezterm & iTerm2 / Tmux & without

umekikazuya:
Fixes #724

- [ n/a ] I have created a discussion that was approved by a maintainer (for new features).

## Problem

In `Select` fields, the cursor remains fixed at the top of the viewport while options scroll up and down. This happens for both `Options()` and `OptionsFunc()`.

Users expect the cursor to move through the visible options, with the viewport only scrolling when the cursor reaches the edges.

## Steps to Reproduce

```go
huh.NewSelect[string]().
    Title("Select an item").
    Height(5).
    Options(
        huh.NewOption("Item 1", "1"),
        huh.NewOption("Item 2", "2"),
        huh.NewOption("Item 3", "3"),
        huh.NewOption("Item 4", "4"),
        huh.NewOption("Item 5", "5"),
        huh.NewOption("Item 6", "6"),
    )
```

* **Before** fix: Pressing down arrow causes "Item 1" to immediately disappear as options scroll, while cursor stays at the top.
* **After** fix: Cursor moves down through visible items. Viewport only scrolls when cursor reaches the bottom.

## Root Cause

In field_select.go:543, `updateViewportHeight()` was setting `s.viewport.YOffset = s.selected` on every update. Since `Update()` calls `updateViewportHeight()` on
every frame, the viewport offset was constantly being reset to match the selected index, creating the "scrolling options" effect.

## Solution

Removed line 543 (`s.viewport.YOffset = s.selected`). The viewport offset is already properly initialized in `selectOption()` and naturally managed by the existing
navigation logic in the `Update()` method.

## Testing

- [x] All existing tests pass 
  - [x] `go test -v ./...`
  - [x] `cd examples && go test -v ./...`
  - [x] `cd spinner && go test -v ./...`
- [x] Tested with `examples/dynamic/dynamic-country`
- [x] Verified both `Options()` and `OptionsFunc()` now behave correctly

### DIFF
--- a/field_select.go
+++ b/field_select.go
@@ -540,7 +540,6 @@ func (s *Select[T]) updateViewportHeight() {
 	}
 
 	s.viewport.Height = max(minHeight, s.height-offset)
-	s.viewport.YOffset = s.selected
 }
 
 func (s *Select[T]) activeStyles() *FieldStyles {


mhelsley (approved):
I think the fix would greatly improve end-users' experience just by reducing the confusion caused by this bug.

Tested on my app and fix confirmed.

ian-howell:
This fix is working for me as well. Would love to see this merged.

umekikazuya:
@aymanbagabas 
Confirmed on current `main` (v2): I can’t reproduce this anymore.  
Cursor behavior is correct for both `Options()` and `OptionsFunc()` (moves through visible items, viewport scrolls only at edges).  
Closing as already fixed.

PR #776, opened 2026-04-28
fix racy light/dark rendering

MichaelMure:
Forms were racingly rendered with light-theme colors regardless of the terminal background, due to three related bugs:
   
  - fix(theme): fix swapped light/dark colors in `ThemeCharm.`  This is a regression that happened in commit 47e59e9
  - fix(form): explicitely request the background color. Before, this was relying on a side effect of `compat.HasDarkBackground` in lipgloss (package level var) triggering the terminal request, which would eventually flow back to the form. This makes it explicit and execute in all cases. Arguably, `compat.HasDarkBackground` can cause problems as it queries `(os.Stdin, os.Stdout)` instead of the real input/output of the form.
  - fix(form): propagate background color message to un-selected groups too

Additionally:
- fix(group): don't propagate events twice. Previously, message other than `tea.KeyPressMsg, tea.PasteMsg` were propagated twice. This also makes the code more obvious.

- [ ] I have created a discussion that was approved by a maintainer (for new features).

Before fix:
image

After fix:
image

### DIFF
--- a/form.go
+++ b/form.go
@@ -521,6 +521,7 @@ func (f *Form) Init() tea.Cmd {
 	}
 
 	cmds = append(cmds, tea.RequestWindowSize)
+	cmds = append(cmds, tea.RequestBackgroundColor)
 	return tea.Sequence(cmds...)
 }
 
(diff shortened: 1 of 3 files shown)


MichaelMure:
After more debugging I realize that this might not be the right fix.

But there **is** a bug: from what I'm seeing, there is a race, at least in `Form` where the `tea.BackgroundColorMsg` get received before or after rendering the form. The ordering is not deterministic between runs of my test program, and the form render with bad colors if `tea.BackgroundColorMsg` get received before rendering (??). This would suggest that the light/dark colors are indeed swapped?

MichaelMure:
I found the real root causes, see my root message.

PR #800, opened 2026-07-07
fix(multiselect): respect limit when toggling a filtered option

upuddu:
When a MultiSelect has a `Limit` and an accepted filter narrows the visible options, toggling could push the selection past the limit: the limit check indexed the full options slice with the filtered cursor (`m.options.val[m.cursor]`) instead of the matched option, so under an active filter it read a different option's `selected` state. This uses the matched option's own index (`i`), matching the line right below it.


kvnloo:
Live check on main ffb6a971 field_multiselect.go:426 still uses m.options.val[m.cursor].selected for the limit gate, then toggles m.options.val[i] where i is the full-list match. After a filter, cursor 0 is Foo in the full list even when the filtered row is Qux, so the gate reads Foo.selected (already true) and lets Qux through.

PR head 98f9ac9a switches the gate to i. TestMultiSelectLimitWithFilter pins it: Foo+Bar at limit 2, filter to Q, toggle, still length 2.

I would merge.

PR #804, opened 2026-08-02
fix(select): show matches hidden above the cursor when filtering

### LINKED ISSUE #669: huh.Select filtering doesn't show all matching options
### Describe the bug

I'm using `huh.Select` to present a list of options to the user. However when I filter down using `/` it doesn't show all of the available options. If I'm specific enough sometimes the option I'm looking for suddenly appears.

Image

When I type `sonnet` initially, only `anthropic/claude-3-sonnet` shows. If I add `-` then delete it suddenly all the `sonnet` models show.

### Setup

Please complete the following information along with version numbers, if applicable.
 - OS: macOS
 - Shell: fish
 - Terminal Emulator: ghostty
 - Terminal Multiplexer: N/A

### To Reproduce

See below.

### Source Code

Pass a list of options into this func:

```go
func showForm(ctx context.Context, models []string) (string, string, error) {
	var (
		question string
		model    = defaultModel
	)

	form := huh.NewForm(
		huh.NewGroup(
			huh.NewInput().
				Title("Enter a task").
				Value(&question),
			huh.NewSelect[string]().
				Title("Select a model").
				Value(&model).
				Height(10).
				Options(huh.NewOptions(models...)...),
		),
	)

	if err := form.RunWithContext(ctx); err != nil {
		log.Fatalf("Failed to run input: %v", err)
	}

	return question, model, nil
}
```

### Expected behavior

Filtering shows all matching results.

### Screenshots

See above.

nf-matt:
Fixes #669.

## The problem

Filtering a `Select` can show fewer matches than actually match. Scrolling up
reveals the rest one at a time, and typing another character then deleting it
also brings them back — which is what makes it look intermittent.

Whether it happens depends on where the cursor was before you filtered, so it
only shows up on lists long enough to scroll.

## Cause

When the filter changes, the previous cursor is carried over and clamped to the
new match count:

```go
s.selected = min(s.selected, len(s.filteredOptions)-1)
```

The viewport is still scrolled to wherever that cursor sat in the unfiltered
list. `ensureCursorVisible()` then runs, but it only ever scrolls the minimum
needed to bring the cursor into view, so it anchors the window on the clamped
cursor and leaves every earlier match above it, out of sight.

It can't self-correct, either: `viewport.SetContent()` runs later during `View`,
so at the point the scroll decision is made the viewport still holds the
previous, longer content and the offset is never clamped against the new length.

## The fix

Move to the first match whenever the filter text *changes*, which re-anchors the
viewport at the top. Navigating within an unchanged filter keeps its position, so
arrow keys behave as before.

Ten options, `WithHeight(6)`, cursor sent to the bottom with `G`, then filtered
with `/ba` — `Bar` and `Baz` both match:

**Before**
```
┃ /ba
┃ > Baz
┃
┃
```

**After**
```
┃ /ba
┃ > Bar
┃   Baz
┃
```

## Testing

Adds `TestSelectFilteringShowsMatchesAboveTheCursor`, which fails without the
change. The existing suite passes.

Note this does change behaviour: filtering now lands on the first match rather
than a clamped previous cursor. That matches `bubbles/list` and survey, and no
existing test relies on the old position — but flagging it explicitly in case
you'd rather preserve the cursor and fix the offset another way.

### DIFF
--- a/field_select.go
+++ b/field_select.go
@@ -329,6 +329,7 @@ func (s *Select[T]) Update(msg tea.Msg) (Model, tea.Cmd) {
 	s.updateViewportSize()
 
 	var cmd tea.Cmd
+	filterBefore := s.filter.Value()
 	if s.filtering {
 		s.filter, cmd = s.filter.Update(msg)
 	}
@@ -492,24 +493,36 @@ func (s *Select[T]) Update(msg tea.Msg) (Model, tea.Cmd) {
 		}
 
 		if s.filtering {
-			s.filteredOptions = s.options.val
-			if s.filter.Value() != "" {
-				s.filteredOptions = nil
-				for _, option := range s.options.val {
-					if s.filterFunc(option.Key) {
-						s.filteredOptions = append(s.filteredOptions, option)
-					}
-				}
-			}
-			if len(s.filteredOptions) > 0 {
-				s.selected = min(s.selected, len(s.filteredOptions)-1)
-			}
+			s.updateFilteredOptions(filterBefore)
 		}
 
 		s.ensureCursorVisible()
 	}
 
 	return s, cmd
+}
+
+func (s *Select[T]) updateFilteredOptions(previousFilter string) {
+	s.filteredOptions = s.options.val
+	if s.filter.Value() != "" {
+		s.filteredOptions = nil
+		for _, option := range s.options.val {
+			if s.filterFunc(option.Key) {
+				s.filteredOptions = append(s.filteredOptions, option)
+			}
+		}
+	}
+	if len(s.filteredOptions) == 0 {
+		return
+	}
+	if s.filter.Value() != previousFilter {
+		// ensureCursorVisible only scrolls the minimum needed, so the offset
+		// has to be reset too or earlier matches stay hidden above the window.
+		s.selected = 0
+		s.viewport.GotoTop()
+		return
+	}
+	s.selected = min(s.selected, len(s.filteredOptions)-1)
 }
 
 func (s *Select[T]) updateValue() {
--- a/huh_test.go
+++ b/huh_test.go
@@ -741,6 +741,31 @@ func TestMultiSelectFiltering(t *testing.T) {
 	})
 }
 
+func TestSelectFilteringShowsMatchesAboveTheCursor(t *testing.T) {
+	// Bar and Baz sit at opposite ends of the list, so filtering for "ba"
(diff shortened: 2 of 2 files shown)


andrinoff:
Hey @nf-matt !

Thank you for your contribution!

Do you mind fixing the lint issue (`nestif`)?

you can @ me afterwards, everything else looks good!

nf-matt:
@andrinoff `nestif` is fixed — pushed just now.

The filtering block was already at complexity 4 before this PR, and the added
`if`/`else` pushed it to 7, so rather than reshuffle the conditionals I moved the
whole block out of `Update` into a `updateFilteredOptions` helper. That matches
how the rest of the field works (`updateValue`, `updateViewportSize`,
`ensureCursorVisible`) and leaves `Update` with a plain `if s.filtering` call.

No behaviour change from the version you looked at — `golangci-lint run ./...`
reports 0 issues locally, and the full suite still passes. I also re-checked that
`TestSelectFilteringShowsMatchesAboveTheCursor` still fails when the cursor reset
is removed, so the regression test is still doing its job after the move.

andrinoff:
Build errors in examples now @nf-matt

nf-matt:
@andrinoff heads up on the red X — lint is green now, and the only failing job is
`build-examples`, which is unrelated to this PR.

It comes from b41d402 (#805), the `charm.land/wish/v2` 2.0.1 → 2.0.2 bump.
`wish/bubbletea` switched its ssh import in that release:

```
v2.0.1  bubbletea/tea.go:10  "github.com/charmbracelet/ssh"
v2.0.2  bubbletea/tea.go:9   "charm.land/ssh"
```

`examples/ssh-form/main.go` still imports `github.com/charmbracelet/ssh`, so
`teaHandler` no longer matches `bubbletea.Handler`. Dependabot labelled it a
patch bump, so it went through unnoticed.

It reproduces on a clean checkout of `main` with nothing from this branch, so it
should be failing on every open PR right now. Pointing the example at
`charm.land/ssh` and running `go mod tidy` builds clean — happy to send that as a
separate PR if it's useful.

andrinoff:
Oh, okay, sorry, will bump

andrinoff:
@nf-matt please, rebase this PR on the latest main, and we should be good-to-merge!

nf-matt:
@andrinoff rebased. thanks!

andrinoff (approved):

PR #808, opened 2026-08-09
fix(group): don't panic on a group with no fields

taciturnaxolotl:
`NewGroup` panics when it is given no fields:

```go
var fields []huh.Field // every value came from a flag, so nothing is left to ask
_ = huh.NewForm(huh.NewGroup(fields...)).Run()
```

```
panic: runtime error: index out of range [0] with length 0
	huh/internal/selector.(*Selector[...]).Selected(...) selector.go:48
	huh.(*Group).getContent(...) group.go:304
	huh.(*Group).Content(...) group.go:382
	huh.(*Group).rawHeight(...) group.go:356
	huh.NewGroup(...) group.go:58
```

`NewGroup` measures its height, which reaches for `selector.Selected()` before checking whether anything is selected.

This is easy to hit from a CLI that builds its form from what the user did not pass on the command line: give every answer as a flag and the slice is empty, so the program crashes on exactly the input that needed no prompting.

### What this changes

- adds `Selector.Empty()` and guards the field accesses that assume a selection, in `Group.Init`, `nextField`, `prevField`, `getContent` and `Footer`, and in `Form.KeyBinds`, `GetFocusedField` and the `nextFieldMsg` handler
- treats a group with no fields as hidden, so a form containing one moves along to the next group rather than showing an empty page
- treats a form with no groups as already complete: `Init` marks it `StateCompleted`, and `Update`, `Errors`, `Help`, `KeyBinds` and `GetFocusedField` return zero values instead of panicking. `NewForm().Init()` panicked the same way before this branch.

`GetFocusedField` and `KeyBinds` can now return nil where they previously panicked; both say so in their doc comments.

### Tests

Three added: an empty group renders nothing and completes, a form with an empty group followed by a real one displays the real one, and a form with no groups at all completes without panicking. `go test ./...` passes.

### DIFF
--- a/form.go
+++ b/form.go
@@ -690,7 +690,7 @@ func (f *Form) styles() FormStyles {
 
 // View renders the form.
 func (f *Form) View() string {
-	if f.quitting {
+	if f.quitting || f.selector.Empty() {
 		return ""
 	}
 
--- a/form.go
+++ b/form.go
@@ -430,17 +430,31 @@ func (f *Form) UpdateFieldPositions() *Form {
 
 // Errors returns the current groups' errors.
 func (f *Form) Errors() []error {
+	if f.selector.Empty() {
+		return nil
+	}
 	return f.selector.Selected().Errors()
 }
 
 // Help returns the current groups' help.
 func (f *Form) Help() help.Model {
+	if f.selector.Empty() {
+		return help.New()
+	}
 	return f.selector.Selected().help
 }
 
 // KeyBinds returns the current fields' keybinds.
+//
+// It returns nil if the form has no groups, or the current group has no fields.
 func (f *Form) KeyBinds() []key.Binding {
+	if f.selector.Empty() {
+		return nil
+	}
 	group := f.selector.Selected()
+	if group.selector.Empty() {
+		return nil
+	}
 	return group.selector.Selected().KeyBinds()
 }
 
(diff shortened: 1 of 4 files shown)


form.go
```diff
@@ -502,7 +505,11 @@ func (f *Form) PrevField() tea.Cmd {
 
 // GetFocusedField returns the focused form field.
 func (f *Form) GetFocusedField() Field {
```
andrinoff, line 520: It can now return nil, which it never could before. Any caller doing `form.GetFocusedField().GetKey()` goes from a panic inside huh to a nil deref in their own code.

group.go
```diff
@@ -393,7 +409,7 @@ func (g *Group) Content() string {
 func (g *Group) Footer() string {
 	var parts []string
 	errors := g.Errors()
-	if g.showHelp && len(errors) <= 0 {
+	if g.showHelp && len(errors) <= 0 && !g.selector.Empty() {
```
andrinoff, line 412: Shouldn't this guard also be in `Header`?

andrinoff (commented):
Also, I think, `f.isGroupHidden(f.selector.Selected())` should also be guarded (will panic now) 

The other comments are just some nitpicks, that would've been nice to have too

taciturnaxolotl:
Thanks for the review! Made a few changes

andrinoff (approved):
LGTM!

